PR #2, opened 2026-01-13
cmr care ported to akd extension

sanzog03:
Summary 📝
Porting a CMR agent created through CARE process using the openai agent platform into the AKD as an extension.

Details
A new CMR agent is created going through the CARE process in openai agent creation platform. The provided agent code is ported to AKD as extension for its easy usability on the AKD backend.  To easily port the generated code with minimal intervention to it, the code is copied to get_response_async method that is extended from the base agent.

Usage
```Python
            from akd_ext.agents.cmr_care_agent import (
                CMRCareAgent,
                CMRCareAgentConfig,
                CMRCareInputSchema,
                CMRCareOutputSchema
            )

            # Initiate Agent
            config = CMRCareAgentConfig()
            agent = CMRCareAgent(config=config, debug=True)

            # Curate input
            test_input = CMRCareInputSchema(input_as_text='Find datasets related to sea surface temperature in the Pacific Ocean')

            # Run agent
            result = await agent.arun(test_input)

            # Display results
            print(f"\nFound {len(result.dataset_concept_ids)} dataset(s):\n")
```

Tested Changes

- [x] Test cases to find the dataset concept ids using the search query. [test with usage](https://github.com/NASA-IMPACT/akd-ext/blob/6f921403d6ec3bf28c23de3bd11381b9f8bf1ab7/tests/test_cmr_care_agent.py)

### DIFF
--- a/akd_ext/agents/_openai_base_agent.py
+++ b/akd_ext/agents/_openai_base_agent.py
@@ -1,0 +1,33 @@
+from agents import HostedMCPTool, Agent, ModelSettings, TResponseInputItem, Runner, RunConfig, trace
+from pydantic import BaseModel, Field
+from openai.types.shared.reasoning import Reasoning
+
+from akd.agents._base import BaseAgent, BaseAgentConfig
+from akd._base import InputSchema, OutputSchema, BaseConfig
+
+class OpenAIBaseAgentConfig(BaseAgentConfig):
+  """Configuration for OpenAI agents based AKD Agents.
+    This configuration extends BaseConfig to provide OpenAI-specific
+    settings for agents built with OpenAI's platform agent builder.
+  """
+  pass
+
+class OpenAIBaseAgent[
+  InSchema: InputSchema,
+  OutSchema: OutputSchema,
+](BaseAgent):
+  """Base class for OpenAI agents.
+  Any agent generated from the openai's platform agent builder should inherit from this class.
+  """
+  input_schema = InputSchema
+  output_schema = OutputSchema
+  
+  def __init__(self,
+      config: OpenAIBaseAgentConfig | None = None,
+      debug: bool = False
+    ) -> None:
+    super().__init__(config=config, debug=debug)
+    self.config = config
+
+  async def _arun(self, params: InSchema, **kwargs) -> OutSchema:
+    raise NotImplementedError("Subclasses must implement this method.")
--- a/akd_ext/agents/cmr_care_agent.py
+++ b/akd_ext/agents/cmr_care_agent.py
@@ -1,0 +1,498 @@
+from akd_ext.agents._openai_base_agent import OpenAIBaseAgent, OpenAIBaseAgentConfig
+from akd._base import InputSchema, OutputSchema
+from agents import HostedMCPTool, Agent, ModelSettings, TResponseInputItem, Runner, RunConfig, trace
+from pydantic import BaseModel, Field
+from openai.types.shared.reasoning import Reasoning
+
+
+# cmr data search agent
+
+class CMRCareAgentConfig(OpenAIBaseAgentConfig):
+  pass
+
+class CMRCareInputSchema(InputSchema):
+  """Input schema for CMR Care Agent."""
+  input_as_text: str = Field(..., description="Input query")
+
+class CMRCareOutputSchema(OutputSchema):
+  """Output schema for CMR Care Agent."""
+  dataset_concept_ids: list[str] = Field(..., description="List of dataset concept IDs")
+
+class CMRCareAgent(OpenAIBaseAgent[CMRCareInputSchema, CMRCareOutputSchema]):
+  input_schema = CMRCareInputSchema
+  output_schema = CMRCareOutputSchema
+  config = CMRCareAgentConfig
+
+  async def _arun(self, params: CMRCareInputSchema) -> CMRCareOutputSchema:
+    result: CMRCareOutputSchema = await self.get_response_async(params)
+    return result
+    
+  async def get_response_async(self, params: CMRCareInputSchema) -> CMRCareOutputSchema:
+    # Tool definitions
+    mcp = HostedMCPTool(tool_config={
+      "type": "mcp",
+      "server_label": "CMR_MCP_Server",
+      "allowed_tools": [
+        "search_collections",
+        "get_granules",
+        "get_collection_metadata"
+      ],
+      "require_approval": "never",
+      "server_description": "CMR MCP server to fetch metadata information including links to download datasets",
+      "server_url": "https://w4hu71445m.execute-api.us-east-1.amazonaws.com/mcp/cmr/mcp"
+    })
+
+    class WorkflowInput(BaseModel):
+      input_as_text: str
+
+    class OutputAgentSchema(BaseModel):
+      dataset_concept_ids: list[str]
+
+    nasa_cmr_data_search_agent = Agent(
+      name="NASA CMR Data Search Agent",
+      instructions="""ROLE
+        You are the NASA Earthdata / CMR Scientific Data Discovery Agent.
+        You are a non-decision-making, human-in-the-loop scientific data discovery assistant whose sole function is to help users discover, organize, and understand NASA Earthdata CMR datasets relevant to Earth science questions.
+        You are not a scientific authority, analyst, or recommender. 
+
+        OBJECTIVE
+        Enable transparent, reproducible, and user-controlled discovery and ranking of NASA Earthdata (CMR) datasets that may answer an Earth science question, including indirect (multi-hop) discovery when direct datasets are insufficient.
+        Your success criteria are:
+        Scientific relevance is reflected only through metadata
+        All assumptions are surfaced and confirmed by the user
+        Users clearly understand why datasets appear
+        No dataset is selected, endorsed, or judged for suitability
+
+        CONTEXT & INPUTS
+        You operate only within Earth science domains:
+        Atmosphere
+        Ocean
+        Land
+        Cryosphere
+        Biosphere
+        Solid Earth
+        You accept:
+        A free-text science question
+        An explicitly selected user expertise level (Intermediate / Advanced)
+        You may use only the following tools and data sources along with the attached context documents: 
+        NASA CMR Search API (REST) — collection discovery only
+        GCMD Keyword Management System (KMS) — vocabulary mapping only
+        Semantic Scholar API — optional, user-approved indirect discovery only 
+        Google Scholar as a last resort.
+        Earthdata Search Web App — link handoff only (no API calls)
+
+        CONSTRAINTS & STYLE RULES
+        Non-Negotiable Guardrails You must never:
+        Recommend, select, or endorse datasets
+        Claim quality, accuracy, uncertainty, or suitability
+        Draw conclusions, trends, causality, or implications
+        Infer or fabricate missing metadata
+        Automate spatial, temporal, or variable assumptions
+        Execute searches without explicit user confirmation
+        Perform downloads or request credentials
+        Operate outside Earth science
+        All missing or ambiguous metadata must be treated as unknown.
+        All indirect (multi-hop) inference requires explicit user approval.
+
+        PROCESS
+        You must follow this canonical reasoning loop exactly:
+        Primary Loop (Direct Discovery First)
+        Interpret the user query into:
+        Phenomenon
+        Explicit variables
+        Expand scientific synonyms (candidate terms only)
+        Clarify (blocking):
+        Variables
+        Spatial bounds
+        Temporal bounds
+        Indirect inference permission (if needed)
+        Map terms → GCMD keywords
+        Translate GCMD concepts → CMR API parameters
+        Search CMR Collections (retrieve multiple candidates)
+        Rank datasets:
+        Primary: metadata relevance
+        Secondary: usage (tie-breaker only)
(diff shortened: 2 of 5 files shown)


NISH1001:
@sanzog03 Thanks for working on this. Couple of things:
- Could you put the PR description in a way it's way more like documentation/elaborate. Eg: What are the major changes. What are the minor changes. IF there's any Usage example for the component, provide that usage section as well.
- The naming conventions: for all the base, i'd encourage to use `_base.py` pattern we use and some of the open source projects follow that convention as well.

[Example PR](https://github.com/NASA-IMPACT/accelerated-discovery/pull/288)

I'd create a PR template that will be easier to fill in next time.

akd_ext/agents/_openai_base_agent.py
```diff
@@ -0,0 +1,33 @@
...
+
+from akd.agents._base import BaseAgent, BaseAgentConfig
+from akd._base import InputSchema, OutputSchema, BaseConfig
+
+class OpenAIBaseAgentConfig(BaseAgentConfig):
+  """Configuration for OpenAI agents based AKD Agents.
+    This configuration extends BaseConfig to provide OpenAI-specific
+    settings for agents built with OpenAI's platform agent builder.
+  """
+  pass
```
NISH1001, line 13: If this doesn't add anything new fields, maybe just removing the config will be better. Or 
Add `tracing_params: dict[str, Any]`? I'd also add `reasoning_effort` as part of the OpenAIBaseAgentConfig and probably `mcp` as well.

```python
class OpenAIBaseAgentConfig(BaseAgentConfig):
  """Configuration for OpenAI agents based AKD Agents.
    This configuration extends BaseConfig to provide OpenAI-specific
    settings for agents built with OpenAI's platform agent builder.
  """
reasoning_effort: Literal["low", "medium", "high",...] = Field(....)
tracing_params: dict[str, Any] | None = Field(default_factory=dict, description="Paramters that can be passed to OpenAI Agents SDK for traceability in the platform API)
```

akd_ext/agents/_openai_base_agent.py
```diff
@@ -0,0 +1,33 @@
...
+  
+  def __init__(self,
+      config: OpenAIBaseAgentConfig | None = None,
+      debug: bool = False
+    ) -> None:
+    super().__init__(config=config, debug=debug)
+    self.config = config
+
+  async def _arun(self, params: InSchema, **kwargs) -> OutSchema:
+    raise NotImplementedError("Subclasses must implement this method.")
```
NISH1001, line 33: We can remove this boilerplate as `akd.agents._base.BaseAgent` already does it. We also don't need to do `self.config=config` because the base meta class already handles this.

What we can do is:
```python

class OpenAIBaseAgent[
  InSchema: InputSchema,
  OutSchema: OutputSchema,
](BaseAgent):
  """Base class for OpenAI agents.
  Any agent generated from the openai's platform agent builder should inherit from this class.
  """

  config = OpenAIBaseAgentConfig
  
  def __init__(self,
      config: OpenAIBaseAgentConfig | None = None,
      debug: bool = False
    ) -> None:
    super().__init__(config=config, debug=debug)
```

akd_ext/agents/_openai_base_agent.py
NISH1001, line 1: Please change the base filename to `akd_ext.agents._base.py`. This will make it consistent in naming convention.

akd_ext/agents/cmr_care_agent.py
```diff
@@ -0,0 +1,498 @@
...
+from akd._base import InputSchema, OutputSchema
+from agents import HostedMCPTool, Agent, ModelSettings, TResponseInputItem, Runner, RunConfig, trace
+from pydantic import BaseModel, Field
+from openai.types.shared.reasoning import Reasoning
+
+
+# cmr data search agent
+
+class CMRCareAgentConfig(OpenAIBaseAgentConfig):
+  pass
```
NISH1001, line 11: Do we need this now? IF it's just a dummy, then we can in fact just remove the config.

Normally, I'd inherit the config if there's extra things that are being added. But in this case `OpenAIBaseAgentConfig` or `BaseAgentConfig` handles it. Or when I want to override some field values. Eg: `reasoning_effort` could be set to `medium` by default for CMR while the base will default to `low`.

Base on the comment in another thread, I think this can take advantage of just overriding `system_prompt` with CARE prompt we have. 

This way, we can make sure Openai agents SDK is properly ported for configurability.

akd_ext/agents/cmr_care_agent.py
```diff
@@ -0,0 +1,498 @@
...
+
+class CMRCareInputSchema(InputSchema):
+  """Input schema for CMR Care Agent."""
+  input_as_text: str = Field(..., description="Input query")
```
NISH1001, line 15: Rename `input_as_text` to just maybe `query` to have consistent input field across everything. And reflect the changes accordingly. This will make consistent for any agents like we have in akd core.

akd_ext/agents/cmr_care_agent.py
```diff
@@ -0,0 +1,498 @@
...
+
+class CMRCareOutputSchema(OutputSchema):
+  """Output schema for CMR Care Agent."""
+  dataset_concept_ids: list[str] = Field(..., description="List of dataset concept IDs")
+
+class CMRCareAgent(OpenAIBaseAgent[CMRCareInputSchema, CMRCareOutputSchema]):
```
NISH1001, line 23: Also don't forget to add an elaborate docstring. Becuase these docstring are used by planner as agent descriptions. So it will help the planner and other guardrail stuff.

akd_ext/agents/cmr_care_agent.py
```diff
@@ -0,0 +1,498 @@
...
+      "server_label": "CMR_MCP_Server",
+      "allowed_tools": [
+        "search_collections",
+        "get_granules",
+        "get_collection_metadata"
+      ],
+      "require_approval": "never",
+      "server_description": "CMR MCP server to fetch metadata information including links to download datasets",
+      "server_url": "https://w4hu71445m.execute-api.us-east-1.amazonaws.com/mcp/cmr/mcp"
+    })
```
NISH1001, line 45: This should be part of the CMR config. If the base config for openai requires some of these, we can add an optional field like mcp to openaibaseagentconfig.

I also encourage to not hardcode param values like server url. Better we can do `os.getenv("CMR_MCP_URL, <default")`

akd_ext/agents/cmr_care_agent.py
```diff
@@ -0,0 +1,498 @@
...
+
+    class WorkflowInput(BaseModel):
+      input_as_text: str
+
+    class OutputAgentSchema(BaseModel):
+      dataset_concept_ids: list[str]
+
+    nasa_cmr_data_search_agent = Agent(
+      name="NASA CMR Data Search Agent",
+      instructions="""ROLE
```
NISH1001, line 56: Move this prompt to CMRAgentConfig.system_prompt.

Since `akd.agents._base.BaseAgentConfig.system_prompt` already exists. So `CMRCareAgentConfig.system_prompt` could be overridden from this. It will make it proper in a way that if we have newer care version for CMR which would be just change in system prompt and instructions TBH. So, `CMRCareAgentConfig.system_prompt` is a proper place.

akd_ext/agents/cmr_care_agent.py
```diff
@@ -0,0 +1,498 @@
...
+        reasoning=Reasoning(
+          effort="low",
+          summary="auto"
+        )
+      )
+    )
+    
+    # Main code entrypoint
+    async def run_workflow(workflow_input: WorkflowInput):
+      with trace("Workshop Data agent"):
```
NISH1001, line 465: the name could come from `tracing_params`

akd_ext/agents/cmr_care_agent.py
```diff
@@ -0,0 +1,498 @@
...
+      "server_url": "https://w4hu71445m.execute-api.us-east-1.amazonaws.com/mcp/cmr/mcp"
+    })
+
+    class WorkflowInput(BaseModel):
+      input_as_text: str
+
+    class OutputAgentSchema(BaseModel):
+      dataset_concept_ids: list[str]
+
+    nasa_cmr_data_search_agent = Agent(
```
NISH1001, line 54: also if we want to properly generalize theopenaibase agent then...the `Agent` object could in fact be abstracted out within the base class as `self.agent` or something.

akd_ext/agents/cmr_care_agent.py
```diff
@@ -0,0 +1,498 @@
...
+          }
+        ]
+        nasa_cmr_data_search_agent_result_temp = await Runner.run(
+          nasa_cmr_data_search_agent,
+          input=[
+            *conversation_history
+          ],
+          run_config=RunConfig(trace_metadata={
+            "__trace_source__": "agent-builder",
+            "workflow_id": "wf_6949ac60e244819082e6ed0bf22ccead09adcd4d789b8c37"
```
NISH1001, line 488: again, this can come from `self.config.tracing_params`

akd_ext/agents/cmr_care_agent.py
```diff
@@ -0,0 +1,498 @@
...
+          })
+        )
+
+        conversation_history.extend([item.to_input_item() for item in output_agent_result_temp.new_items])
```
NISH1001, line 508: I think we should make use of conversation tracking we have in`BaseAgent`. Since `memory` ([see](https://github.com/NASA-IMPACT/accelerated-discovery/blob/develop/akd/agents/_base.py#L131)) is abstracted in base...we can have `OpenAIBaseAgent.memory` as list of dict. Looking a tit. IT does seems like it's a list.

NISH1001 (changes requested):
Thank you for quickly having this PR. I have added a detail feedback, particularly how we want to design the openaisdk porting.

pyproject.toml
```diff
@@ -9,6 +9,7 @@ dependencies = [
     "openai-agents>=0.1.0",
     "openai>=1.0.0",
     "pydantic>=2.0.0",
+    "deepeval>=3.0.0",
```
NISH1001, line 12: why do we need deepeval in akd-ext? for what purpose?

NISH1001 (changes requested):

pyproject.toml
NISH1001: deepeval is akd core dependency which should be installed with "ml" if needed.

`uv sync --extra ml`

NISH1001: I am trying to avoid large dependency in akd extension unless needed. In future we will have `akd-core` as separate repo which will again make the extension minimal in itself. We only need deepeval if you're playing around locally with something like `RiskAgent`

uv.lock
```diff
@@ -0,0 +1,4632 @@
...
+    { name = "wikipedia" },
+]
+
+[[package]]
+name = "akd-ext"
+version = "0.0.1"
+source = { editable = "." }
+dependencies = [
+    { name = "akd" },
+    { name = "deepeval" },
```
NISH1001, line 179: no need to install deepeval for akd-ext.

uv.lock
```diff
@@ -0,0 +1,4632 @@
...
+    { name = "openai" },
+    { name = "openai-agents" },
+    { name = "pydantic" },
+]
+
+[package.metadata]
+requires-dist = [
+    { name = "akd", git = "https://github.com/NASA-IMPACT/accelerated-discovery.git?rev=develop" },
+    { name = "deepeval", specifier = ">=3.0.0" },
```
NISH1001, line 188: same.

NISH1001 (changes requested):

pyproject.toml
NISH1001: Okay i think this is akd-core problem in architecture. i am resolving that in akd-core so that akd-ext doesn't have any heavy dependency. TBD

NISH1001: This PR in akd-core solves this: https://github.com/NASA-IMPACT/accelerated-discovery/pull/311

uv.lock
NISH1001: resolved via akd-core fix: https://github.com/NASA-IMPACT/accelerated-discovery/pull/311

uv.lock
NISH1001: Please remove deepeval now

akd_ext/agents/_openai_base_agent.py
NISH1001: resolved by #7

akd_ext/agents/_openai_base_agent.py
NISH1001: resolved by #7

akd_ext/agents/cmr_care_agent.py
NISH1001: resolved by #7

akd_ext/agents/cmr_care_agent.py
NISH1001: resolved by #7

akd_ext/agents/cmr_care_agent.py
NISH1001: resolved by #7

NISH1001:
Closing this in favor of #9

akd_ext/agents/cmr_care_agent.py
NISH1001: fixes in #9

Issue #4, opened 2026-01-14
Extend Code Repo Search Tool


NISH1001:
Some extra work is needed to extend existing search tool.

The tool is already there in akd-core:
https://github.com/NASA-IMPACT/accelerated-discovery/blob/develop/akd/tools/search/code_search.py#L629

The tool already follows akd design principle.

Extra work includes:
- Incorporating reliability score. Get the formula from Ankur
- For this to happen, use GH API to get stars, commits, forks, metadata
- I think the output schema needs update to incorporate the score.

@sanzog03 .

sanzog03:
Pr: https://github.com/NASA-IMPACT/akd-ext/pull/11 @NISH1001

PR #11, opened 2026-01-20
Task/akd repository search tool

sanzog03:
Summary 📝
This PR implements #4 a Repository Search Tool that extends the existing SDECodeSearchTool to enrich code search results with GitHub repository metadata and reliability scoring. The tool fetches repository statistics (stars, forks, issues, pull requests) and computes a reliability score for each search result, enabling better assessment of code quality and repository health.

The tool supports optional GitHub authentication via access tokens to avoid rate limiting and access private repositories.

Details
1. RepositorySearchTool: Extended SDECodeSearchTool to add GitHub metadata enrichment
- Inherits base code search functionality
- Fetches repository metadata using PyGithub library
- Computes reliability scores based on repository metrics
2. RepositoryMetadata: Dataclass containing GitHub repository statistics
- Stars count
- Forks count
- Open issues count
- Open pull requests count
- Closed pull requests count
3. RepositorySearchToolSearchResultItem: Extended 
- SearchResultItem with additional fields
4. reliability_score: Computed score based on repository health
5. Configuration & Authentication via RepositorySearchToolConfig: Extended configuration schema
- Added optional access_token field for GitHub authentication
- Inherits all base code search configuration options
- Uses unauthenticated access by default
- Supports GitHub token authentication when access_token is provided
- Prevents rate limiting for high-volume searches
6. Schema Compatibility Fix on Output Schema Workaround: Implemented temporary fix for Pydantic validation error
- Parent class uses self.output_schema internally during execution
- Temporarily swaps to parent's schema before calling super()._arun()
- Restores custom schema after parent execution completes
- Marked with TODO for future refactoring
- Usage Section The 
RepositorySearchTool can be used to search for code while automatically enriching results with GitHub repository metadata:

Basic Usage (No Authentication)
```python
from akd_ext.tools.repository_search_tool import (
    RepositorySearchTool,
    RepositorySearchToolConfig,
    RepositorySearchToolInputSchema
)

# Create tool with default config
config = RepositorySearchToolConfig(page_size=10)
tool = RepositorySearchTool(config=config)
# Execute search
result = await tool._arun(
    RepositorySearchToolInputSchema(queries=["machine learning pipeline"])
)
# Access enriched results
for item in result.results:
    print(f"Repository: {item.url}")
    print(f"Stars: {item.repository_metadata.stars}")
    print(f"Forks: {item.repository_metadata.forks}")
    print(f"Reliability Score: {item.reliability_score}")

```
Authenticated Usage (With GitHub Token)
```python
# Create tool with authentication
config = RepositorySearchToolConfig(
    page_size=10,
    access_token="ghp_your_github_token_here"  # Avoids rate limiting
)
tool = RepositorySearchTool(config=config)
# Execute search with higher rate limits
result = await tool._arun(
    RepositorySearchToolInputSchema(queries=["data processing"])
)
```
Result Structure
Each search result includes:

```python
RepositorySearchToolSearchResultItem(
    # Standard SearchResultItem fields
    query="...",
    url="https://github.com/owner/repo",
    title="...",
    snippet="...",
    
    # Enhanced fields
    reliability_score=1.0,  # Computed score (placeholder implementation)
    repository_metadata=RepositoryMetadata(
        stars=1234,
        forks=567,
        open_issues=89,
        pulls=12,
        closed_pulls=345
    )
)
```
Configuration Options
All SDECodeSearchToolConfig options are supported, plus:

access_token (Optional[str]): GitHub personal access token for authenticated requests

Checks
 Tested Changes - Script runs successfully with test query

### DIFF
--- a/akd_ext/tools/code_search/repository_search.py
+++ b/akd_ext/tools/code_search/repository_search.py
@@ -10,6 +10,7 @@ from akd.tools.search.code_search import (
 )
 from akd.structures import SearchResultItem
 from akd_ext.tools.code_search.utils import RepositoryMetadata, fetch_github_metadata, calculate_reliability_score
+from loguru import logger
 
 
 class RepositorySearchResultItem(SearchResultItem):
@@ -112,13 +113,10 @@ if __name__ == "__main__":
     import asyncio
     import sys
 
-    async def main():
-        config = RepositorySearchToolConfig(page_size=2)
-        query = "indus pipeline code"
-        if len(sys.argv) > 1:
-            query = sys.argv[1]
-        tool = RepositorySearchTool(config=config)
-        result = await tool.arun(RepositorySearchToolInputSchema(queries=[query]))
-        print(result.model_dump())
-
-    asyncio.run(main())
+    config = RepositorySearchToolConfig(page_size=2)
+    query = "indus pipeline code"
+    if len(sys.argv) > 1:
+        query = sys.argv[1]
+    tool = RepositorySearchTool(config=config)
+    result = asyncio.run(tool.arun(RepositorySearchToolInputSchema(queries=[query])))
+    logger.info(result.model_dump())
--- a/akd_ext/tools/code_search/repository_search.py
+++ b/akd_ext/tools/code_search/repository_search.py
@@ -71,7 +71,11 @@ class RepositorySearchTool(SDECodeSearchTool):
       # collect necessary metadata
       repository_metadata: RepositoryMetadata = await fetch_github_metadata(repo_name, self.config.access_token)
       reliability_score: int | float = calculate_reliability_score(repository_metadata)
-      repository_search_result.results.append(RepositorySearchResultItem(**repository.model_dump(), repository_metadata=repository_metadata, reliability_score=reliability_score))
+      repository_search_result.results.append(RepositorySearchResultItem(**{
+        **repository.model_dump(),
+        "repository_metadata": repository_metadata,
+        "reliability_score": reliability_score
+        }))
     return repository_search_result
 
 if __name__ == "__main__":
--- a/akd_ext/tools/code_search/repository_search.py
+++ b/akd_ext/tools/code_search/repository_search.py
@@ -70,8 +70,23 @@ class RepositorySearchToolConfig(SDECodeSearchToolConfig):
 # Tool implementation
 class RepositorySearchTool(SDECodeSearchTool):
     """
-    Search for relevant code that exists in the github repository. It extends the code search tool and adds github repository metadata and reliability score to each search result item.
-    Note: The code search tool uses SDE API.
+    Search for relevant code and implementations within specialized science repositories.
+
+    This tool performs a targeted search across curated scientific codebases to find
+    relevant repositories with README. It enriches the search results with
+    GitHub metadata such as stars, forks, and development activity, which are then
+    used to compute a reliability score for each item.
+
+    The reliability score (0-100) is a weighted average of repository maturity, activity, and community trust.
+
+    The formula: Score = (Age * 0.20) + (Activity * 0.25) + (Stars * 0.25) + (Forks * 0.15) + (History * 0.15)
+
+    How components are calculated:
+      - Age (20%): Higher for older repos; reaches 100% after 4 years.
+      - Activity (25%): Starts at 100% and drops to 0% if the repo hasn't been updated in a year.
+      - Stars (25%): Logarithmic scale where ~1,000 stars = 100%.
+      - Forks (15%): Logarithmic scale where ~500 forks = 100%.
+      - History (15%): Based on the span between the first commit and now; reaches 100% after 4 years.
     """
 
     input_schema = RepositorySearchToolInputSchema
--- a/akd_ext/tools/code_search/repository_search.py
+++ b/akd_ext/tools/code_search/repository_search.py
@@ -1,5 +1,6 @@
 import os
 from dataclasses import dataclass, field
+from urllib.parse import urlparse
 from pydantic import Field, BaseModel
 from github import Github, Auth
 from akd.tools._base import BaseTool
@@ -62,9 +63,10 @@ class RepositorySearchTool(SDECodeSearchTool):
       url: str = str(repository.url)
       if not url:
         continue
-      # Github library to get repository metadata
-      parts = url.rstrip('/').split('github.com/')[-1].split('/')
-      owner, repo = parts[0], parts[1]
+      # Parse URL to extract owner/repo from github url
+      parsed_url = urlparse(url)
+      path_parts = parsed_url.path.strip('/').split('/')
+      owner, repo = path_parts[0], path_parts[1]
       repo_name = f"{owner}/{repo}"
       # collect necessary metadata
       repository_metadata: RepositoryMetadata = await fetch_github_metadata(repo_name, self.config.access_token)
--- a/akd_ext/tools/code_search/repository_search.py
+++ b/akd_ext/tools/code_search/repository_search.py
@@ -3,7 +3,6 @@ from pydantic import Field, BaseModel
 from github import Github, Auth
 from akd.tools._base import BaseTool
 from akd.tools.search.code_search import SDECodeSearchTool, SDECodeSearchToolConfig, CodeSearchToolInputSchema, CodeSearchToolOutputSchema
-from akd.tools.search._base import SearchToolOutputSchema
 from akd.structures import SearchResultItem
 
 class RepositoryMetadata(BaseModel):
@@ -29,7 +28,7 @@ class RepositorySearchToolInputSchema(CodeSearchToolInputSchema):
     Input query for the repository search tool. Its a text based query that initializes the relevant code search tool.
   """
 
-class RepositorySearchToolOutputSchema(SearchToolOutputSchema):
+class RepositorySearchToolOutputSchema(CodeSearchToolOutputSchema):
   """
     Output schema for the repository search tool.
   """
--- a/akd_ext/tools/code_search/repository_search.py
+++ b/akd_ext/tools/code_search/repository_search.py
@@ -1,4 +1,3 @@
-from typing import Optional
 from dataclasses import dataclass, field
 from pydantic import Field, BaseModel
 from github import Github, Auth
@@ -17,7 +16,7 @@ class RepositoryMetadata(BaseModel):
   pulls: int = Field(default=0, description="Number of open pull requests on the repository.")
   closed_pulls: int = Field(default=0, description="Number of closed pull requests on the repository.")
 
-class RepositorySearchToolSearchResultItem(SearchResultItem):
+class RepositorySearchResultItem(SearchResultItem):
   """
     Search result item with added github repository metadata and computed reliability score.
   """
@@ -34,7 +33,7 @@ class RepositorySearchToolOutputSchema(SearchToolOutputSchema):
   """
     Output schema for the repository search tool.
   """
-  results: list[RepositorySearchToolSearchResultItem] = Field(
+  results: list[RepositorySearchResultItem] = Field(
     ...,
     description="List of search result items with added github repository metadata and computed reliability score.",
   )
@@ -44,7 +43,7 @@ class RepositorySearchToolConfig(SDECodeSearchToolConfig):
   """
     Config schema for the repository search tool.
   """
-  access_token: Optional[str] = Field(default=None, description="GitHub access token")
+  access_token: str | None = Field(default=None, description="GitHub access token")
 
 # Tool implementation
 class RepositorySearchTool(SDECodeSearchTool):
@@ -89,7 +88,7 @@ class RepositorySearchTool(SDECodeSearchTool):
         repository_metadata.pulls = repo.get_pulls(state='open', sort='created', base='master').totalCount
         repository_metadata.closed_pulls = repo.get_pulls(state='closed', sort='created', base='master').totalCount
       reliability_score = self._get_reliability_score(repository_metadata)
-      repository_search_result.results.append(RepositorySearchToolSearchResultItem(**repository.model_dump(), repository_metadata=repository_metadata, reliability_score=reliability_score))
+      repository_search_result.results.append(RepositorySearchResultItem(**repository.model_dump(), repository_metadata=repository_metadata, reliability_score=reliability_score))
     return repository_search_result
 
   def _get_reliability_score(self, repository_metadata: RepositoryMetadata) -> float:
--- a/akd_ext/tools/code_search/repository_search.py
+++ b/akd_ext/tools/code_search/repository_search.py
@@ -74,7 +74,7 @@ class RepositorySearchTool(SDECodeSearchTool):
     owner, repo = path_parts[0], path_parts[1]
     repo_name = f"{owner}/{repo}"
     repository_metadata: RepositoryMetadata = await fetch_github_metadata(repo_name, self.config.access_token)
-    reliability_score: int | float = calculate_reliability_score(repository_metadata)
+    reliability_score: float = calculate_reliability_score(repository_metadata)
     return RepositorySearchResultItem(**{
       **repository_item.model_dump(),
       "repository_metadata": repository_metadata,
--- a/akd_ext/tools/code_search/repository_search.py
+++ b/akd_ext/tools/code_search/repository_search.py
@@ -73,7 +73,7 @@ class RepositorySearchTool(SDECodeSearchTool):
     Search for relevant code and implementations within specialized science repositories.
 
     This tool performs a targeted search across curated scientific codebases to find
-    relevant repositories with README. It enriches the search results with
+    relevant GitHub repositories with README. It enriches the search results with
     GitHub metadata such as stars, forks, and development activity, which are then
     used to compute a reliability score for each item.
 
--- a/akd_ext/tools/code_search/repository_search.py
+++ b/akd_ext/tools/code_search/repository_search.py
@@ -1,95 +1,124 @@
 import os
 import asyncio
-from dataclasses import dataclass, field
-from pydantic import Field
-from typing import List
+from pydantic import Field, model_validator
 from urllib.parse import urlparse
-from github import Github, Auth
-from akd.tools._base import BaseTool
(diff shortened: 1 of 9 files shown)


akd_ext/tools/repository_search_tool.py
```diff
@@ -0,0 +1,109 @@
...
+from akd.structures import SearchResultItem
+
+# Tool specific data
+@dataclass
+class RepositoryMetadata:
+  stars: int = 0
+  forks: int = 0
+  open_issues: int = 0
+  pulls: int = 0
+  closed_pulls: int = 0
```
NISH1001: Convert this to pydantic please.

```python
class RepositoryMetadata:
stars: int = Field(default=0, description=...)
....
```

akd_ext/tools/repository_search_tool.py
```diff
@@ -0,0 +1,109 @@
...
+
+class RepositorySearchToolSearchResultItem(SearchResultItem):
```
NISH1001: Rename this to `RepositorySearchResultItem`.

akd_ext/tools/repository_search_tool.py
```diff
@@ -0,0 +1,109 @@
...
+    ...,
+    description="List of search result items with added github repository metadata and computed reliability score.",
+  )
+
+# Tool config schema
+class RepositorySearchToolConfig(SDECodeSearchToolConfig):
+  """
+    Config schema for the repository search tool.
+  """
+  access_token: Optional[str] = Field(default=None, description="GitHub access token")
```
NISH1001: Use modern python 3.12+ style
```python
access_token: str | None = FIeld(defaut=None, .......)
```

akd_ext/tools/code_search/repository_search.py
NISH1001, line 1: Change the namepsace to `tools/code_search/repository_search.py`. I recommend not using the suffix for filename because it'already implied by namespace itself.

akd_ext/tools/repository_search_tool.py
```diff
@@ -0,0 +1,109 @@
...
+  reliability_score: float = Field(default=1.0, description="Computed reliability score based on github repository metadata.")
+  repository_metadata: RepositoryMetadata = Field(default_factory=RepositoryMetadata, description="Github repository metadata. includes number of stars, forks, open issues, open pull requests, and closed pull requests.")
+
+# Tool input and output schemas
+class RepositorySearchToolInputSchema(CodeSearchToolInputSchema):
+  """
+    Input query for the repository search tool. Its a text based query that initializes the relevant code search tool.
+  """
+
+class RepositorySearchToolOutputSchema(SearchToolOutputSchema):
```
NISH1001: Inherit from`CodeSearchToolOutputSchema`, which is just one step up in inheritance hierarchy...

akd_ext/tools/code_search/repository_search.py
```diff
@@ -0,0 +1,109 @@
...
+  
+  async def _arun(self, params: RepositorySearchToolInputSchema) -> RepositorySearchToolOutputSchema:
+    # Temporarily use parent's output schema to avoid validation error caused by super()._arun(params) + parents output schema
+    # TODO: better fix for this issue
+    original_output_schema = self.output_schema
+    self.output_schema = CodeSearchToolOutputSchema
+    try:
+      search_result: CodeSearchToolOutputSchema = await super()._arun(params)
+    finally:
+      self.output_schema = original_output_schema
```
NISH1001: This would be fixed if we just use `SearchResultItem` directly in the list instead of `RepositorySearchToolSearchResultItem`. I will check what can be done at akd-core level. But for now can you revert to using `searchresultitem` and put the repo metadata inside extra there?

akd_ext/tools/code_search/repository_search.py
```diff
@@ -0,0 +1,109 @@
...
+      auth = None
+      if self.config.access_token:
+        auth = Auth.Token(self.config.access_token)
+      # collect necessary metadata
+      repository_metadata: RepositoryMetadata = RepositoryMetadata()
+      with Github(auth=auth) as g:
+        repo = g.get_repo(repo_name)
+        repository_metadata.stars = repo.stargazers_count
+        repository_metadata.forks = repo.forks_count
+        repository_metadata.open_issues = repo.get_issues(state='open').totalCount
```
NISH1001: Move this logic to a utility. potentially in `akd_ext/tools/code_search/utils.py` please. This way anyone can use it and it's functional. And also parallelize the metadata fetching with asyncio.gather

```python
async def fetch_github_metadata(access_token: str, repo_url: ...) -> dict[str, int|float]:
...do the magic...
return dict
```

Then

```python

# note: the fetch is async, and we're not doing "await" here. This creates a task list
tasks = [ fetch_github_metadata(acesstoken=..., url=result.url) for result in results]

# this will concurrently handle things
metadata_list = asyncio.gather(*tasks)

then iterate
```

akd_ext/tools/code_search/repository_search.py
```diff
@@ -0,0 +1,109 @@
...
+        repository_metadata.pulls = repo.get_pulls(state='open', sort='created', base='master').totalCount
+        repository_metadata.closed_pulls = repo.get_pulls(state='closed', sort='created', base='master').totalCount
+      reliability_score = self._get_reliability_score(repository_metadata)
+      repository_search_result.results.append(RepositorySearchToolSearchResultItem(**repository.model_dump(), repository_metadata=repository_metadata, reliability_score=reliability_score))
+    return repository_search_result
+
+  def _get_reliability_score(self, repository_metadata: RepositoryMetadata) -> float:
+    # TODO: use reliability score calculation formula
+    return 1.0
```
NISH1001: Move this logic to utils.py as well so that it's reusable as an independent function.

pyproject.toml
```diff
@@ -16,8 +17,9 @@ classifiers = [
 ]
 requires-python = ">=3.12"
 dependencies = [
-    "akd @ git+https://github.com/NASA-IMPACT/accelerated-discovery.git@develop",
+    "akd[ml] @ git+https://github.com/NASA-IMPACT/accelerated-discovery.git@develop",
```
NISH1001: Don't put this as dependencies please. It's not how it is. Remember: akd-ext will be deployed with serverless with limited size. I am trying to make akd-ext lightweight.

There's ipython issue. So don't add ml as dependencies. Any dependencies for akd core will be fixed in akd-core.

NISH1001: And update the uv.loc after removing "ml" namespace. we don't need ml from akd-core. If there are things needed, we will fix in akd-core dependency graphs.

NISH1001: The only deps heavy for code search is sentence transformer which I think is for reranking. i can fix that at akd-core by lazy loading that module without having to bloat akd-ext core dependencies.

akd_ext/tools/code_search/repository_search.py
NISH1001: Meanwhile ia m investigating a fix in akd-core for this.

NISH1001 (changes requested):

akd_ext/tools/code_search/repository_search.py
NISH1001: Another way to fix this could be by adding a validator to reposearchreusltitem.:

```python

 class RepositorySearchToolSearchResultItem(SearchResultItem):                                                       
     reliability_score: float = Field(......)                                                                                              
     repository_metadata: RepositoryMetadata = Field(.....)                                                                                              
                                                                                                                     
     @model_validator(mode='before')                                                                                 
     @classmethod                                                                                                    
     def convert_parent_instance(cls, data):                                                                         
         """Convert SearchResultItem objects to dicts so Pydantic can create child instances."""                     
         if isinstance(data, SearchResultItem) and not isinstance(data, cls):                                        
             return data.model_dump()                                                                                
         return data    
```

The reason we see the error is: the tool when doing `_arun(...)` for parent, creating `SearchResultItem` but the current class codesearhctool has this new subclass. so pydantic can't convert object of parent class to that of subclass. This should work.

NISH1001:
@sanzog03 i have reviewe the pr. Also few notes:
- The branch name follows `feature/` convention for any new feature. 
- THink of repo as a package itself which will be published. So `task/` feels weirder.

akd_ext/tools/code_search/repository_search.py
```diff
@@ -0,0 +1,109 @@
...
+      self.output_schema = original_output_schema
+
+    repository_search_result: RepositorySearchToolOutputSchema = RepositorySearchToolOutputSchema(results=[], extra=search_result.extra)
+    # add repository metadata to each result
+    for repository in search_result.results:
+      url: str = str(repository.url)
+      if not url:
+        continue
+      # Github library to get repository metadata
+      parts = url.rstrip('/').split('github.com/')[-1].split('/')
```
NISH1001: Also instead of manually splitting, we should probably use better way like pathlib or urlparse

NISH1001 (changes requested):

akd_ext/tools/repository_search_tool.py
```diff
@@ -0,0 +1,109 @@
...
+      repository_metadata: RepositoryMetadata = RepositoryMetadata()
+      with Github(auth=auth) as g:
+        repo = g.get_repo(repo_name)
+        repository_metadata.stars = repo.stargazers_count
+        repository_metadata.forks = repo.forks_count
+        repository_metadata.open_issues = repo.get_issues(state='open').totalCount
+        repository_metadata.pulls = repo.get_pulls(state='open', sort='created', base='master').totalCount
+        repository_metadata.closed_pulls = repo.get_pulls(state='closed', sort='created', base='master').totalCount
+      reliability_score = self._get_reliability_score(repository_metadata)
+      repository_search_result.results.append(RepositorySearchToolSearchResultItem(**repository.model_dump(), repository_metadata=repository_metadata, reliability_score=reliability_score))
```
NISH1001: If we're going with model_validator route, then this could be

Option 1:

```
updated_item = repository.model_copy(update={                                                                       
     "repository_metadata": repository_metadata,                                                                     
     "reliability_score": reliability_score                                                                          
 })
repository_search_result.results.append(updated_item)
```

Option 2:
```python
repository_search_result.results.append(RepositorySearchToolSearchResultItem(**{                                    
     **repository.model_dump(),                                                                                      
     "repository_metadata": repository_metadata,                                                                     
     "reliability_score": reliability_score                                                                          
 })) 
```

Again, this is assuming model_validator fixes the whole issue with temporary copy you had done.

NISH1001 (changes requested):

tests/tools/code_search/test_repository_search.py
```diff
@@ -0,0 +1,40 @@
...
+)
+async def test_repository_search_tool(query: str):
+    """Test Repository Search Tool functionality.
+
+    Args:
+        query: Code search query to test
+    """
+    config = RepositorySearchToolConfig(page_size=2)
+    tool = RepositorySearchTool(config=config)
+    result = await tool._arun(RepositorySearchToolInputSchema(queries=[query]))
```
NISH1001: Always use `arun` which is the public facing interface

akd_ext/tools/code_search/repository_search.py
```diff
@@ -0,0 +1,109 @@
...
+
+if __name__ == "__main__":
+  import asyncio
+  
+  async def main():
+    config = RepositorySearchToolConfig(
+      page_size = 2     
+    )
+    tool = RepositorySearchTool(config=config)
+    result = await tool._arun(RepositorySearchToolInputSchema(queries=["indus pipeline code"]))
```
NISH1001: Always use `arun` which is the public facing interface

NISH1001 (changes requested):

akd_ext/tools/repository_search_tool.py
NISH1001: I just confirmed. this fixes the issue with the hacky to do you had.

akd_ext/tools/code_search/repository_search.py
NISH1001: I tested myself. ALso see my other comment down.

akd_ext/tools/repository_search_tool.py
sanzog03: changed!

NISH1001:
@sanzog03 few things:
- THe model validator fixes the hacky solution you had (see both of the threads)
- Like mentioned, we need to make the metadata fetching part concurrent using asycnio gather. This will speed things up drastically. Also add `max_concurrency` and default to 5. use `asyncio.semaphore` to limit concurrency.

akd_ext/tools/repository_search_tool.py
sanzog03: changed!

akd_ext/tools/repository_search_tool.py
sanzog03: changed!

akd_ext/tools/code_search/repository_search.py
sanzog03: changed!

akd_ext/tools/code_search/repository_search.py
```diff
@@ -0,0 +1,109 @@
...
+  """
+    Config schema for the repository search tool.
+  """
+  access_token: Optional[str] = Field(default=None, description="GitHub access token")
+
+# Tool implementation
+class RepositorySearchTool(SDECodeSearchTool):
+  """
+  Search for relevant code that exists in the github repository. It extends the code search tool and adds github repository metadata and reliability score to each search result item.
+  Note: The code search tool uses SDE API.
```
NISH1001: Refactor this docstring. Avoid using SDE in the docstirng for now because it might confuse the agent for another tool simran is making. Can you expand docstring. Say escpially this is using science repositories.

NISH1001 (changes requested):

pyproject.toml
NISH1001: If this is needed, then I need to change akd-core itself because I was assuming akd-ext would be thin wrapper without ML heavy stuff for time being.

akd_ext/tools/repository_search_tool.py
sanzog03: changed!

akd_ext/tools/code_search/repository_search.py
NISH1001: FOr concurrnecy. we could try `akd.utils.RateLimiter`?

```python
import httpx
from akd.utils import RateLimiter

class MyApiClient:
    def __init__(self, base_url: str, rate_limiter: RateLimiter | None = None):
        self.base_url = base_url
        self.rate_limiter = rate_limiter or RateLimiter(max_calls_per_second=2.0)

    async def get(self, path: str):
        # Gate the request
        await self.rate_limiter.acquire()

        async with httpx.AsyncClient(base_url=self.base_url) as client:
            r = await client.get(path)
            r.raise_for_status()
            return r.json()
```

I will let you decide how to go on about rate limiting or for now we can just do asyncio.gather without any semaphore / locks

akd_ext/tools/code_search/repository_search.py
```diff
@@ -0,0 +1,108 @@
...
+    ...,
+    description="List of search result items with added github repository metadata and computed reliability score.",
+  )
+
+# Tool config schema
+class RepositorySearchToolConfig(SDECodeSearchToolConfig):
+  """
+    Config schema for the repository search tool.
+  """
+  access_token: str | None = Field(default=None, description="GitHub access token")
```
NISH1001: also get from env variable default nad default to None.


`os.getenv("GITHUB_ACCESS_TOKEN", None)`

NISH1001 (changes requested):

akd_ext/tools/code_search/repository_search.py
sanzog03: changed!

akd_ext/tools/code_search/repository_search.py
sanzog03: changed!

akd_ext/tools/code_search/repository_search.py
sanzog03: changed!

tests/tools/code_search/test_repository_search.py
sanzog03: changed!

sanzog03:
> @sanzog03 i have reviewe the pr. Also few notes:
> 
> * The branch name follows `feature/` convention for any new feature.
> * THink of repo as a package itself which will be published. So `task/` feels weirder.

will have this in the upcoming prs.

akd_ext/tools/code_search/repository_search.py
sanzog03: changed!

akd_ext/tools/code_search/repository_search.py
sanzog03: @NISH1001 The PyGithub library (is handling the HTTP **request**) along with github will have the ratelimit. Do we need an additional rate limit on top?

akd_ext/tools/repository_search_tool.py
sanzog03: changed!

akd_ext/tools/code_search/repository_search.py
sanzog03: changed!

akd_ext/tools/code_search/repository_search.py
sanzog03: changed!

pyproject.toml
sanzog03: changed!

akd_ext/tools/code_search/repository_search.py
```diff
@@ -0,0 +1,137 @@
...
+    access_token: str | None = Field(default=os.getenv("GITHUB_ACCESS_TOKEN", None), description="GitHub access token")
+
+
+# Tool implementation
+class RepositorySearchTool(SDECodeSearchTool):
+    """
+    Search for relevant code and implementations within specialized science repositories.
+
+    This tool performs a targeted search across curated scientific codebases to find
+    relevant repositories with README. It enriches the search results with
```
NISH1001: specify "relevant GitHub repositories" so that we are explicit about this being GitHub only thing

sanzog03: changed!

NISH1001 (approved):

PR #13, opened 2026-01-20
feat: Add akd_ext/mcp module for FastMCP integration

rohit-sahoo:
## Summary

Add MCP module (`akd_ext/mcp/`) that converts akd BaseTools to FastMCP-compatible functions, enabling tools to be exposed via Model Context Protocol (MCP).

**Core components:**
- `tool_converter` - Converts BaseTool to async MCP function using Signature-based approach
- `@mcp_tool` decorator - Auto-registers tools at import time
- `MCPToolRegistry` - Singleton registry (follows akd-core pattern)
- `server.py` - FastMCP server with auto-discovery + manual registration option

### Dependencies:
- `fastmcp>=2.0.0`
- `pydantic>=2.0.0`

### Option 1: Decorator (recommended)
```python
@mcp_tool
class MyTool(BaseTool):
    ...
```

### Option 2: Manual registration
```python
# In server.py
MANUAL_TOOLS: list[type[BaseTool]] = [InternalTool]
```

## Testing

- Unit tests for converter, registry, decorator
- Can be deployed to FastMCP Cloud - Entrypoint: akd_ext/mcp/server.py:mcp

### Note:
- The _create_wrapper function doesn't do any I/O, it just builds and returns a function. Only the inner `async def _async_wrapper()` wrapper needs to be async because `tool.arun()` is async. 
```python
def _create_wrapper():           # sync - just creates the function
    async def _async_wrapper():  # async - actual tool execution
        result = await tool.arun(params)
        return result
    return _async_wrapper
```
- We use `inspect.Signature` and `bind()` instead of string-based `exec()` 
- Follows `akd.planner.registry.AgentRegistry` pattern which provides single source of truth and prevents duplicate registrations


akd_ext/mcp/mcp/converter.py
```diff
@@ -0,0 +1,85 @@
+"""MCP Tool Converter - Converts akd BaseTool to FastMCP compatible functions."""
+
+from collections.abc import Callable
+from inspect import Signature, Parameter
+
+from fastmcp import FastMCP
+from akd.tools._base import BaseTool
+
+
+def tool_converter(tool: BaseTool) -> Callable:
```
NISH1001: if the wrapper (see comment down) is async, the return type would change.

Something like:

```python
from typing import Awaitable
from collections.abc import Callable
def tool_converter(tool: BaseTool) -> Callable[..., Awaitable[Any]]:
...
....
```

akd_ext/mcp/mcp/converter.py
```diff
@@ -0,0 +1,85 @@
...
+    # Set function metadata
+    mcp_tool_wrapper.__name__ = tool_name
+    mcp_tool_wrapper.__doc__ = tool_description
+    mcp_tool_wrapper.__signature__ = wrapper_sig
+    mcp_tool_wrapper.__annotations__ = annotations
+
+    return mcp_tool_wrapper
+
+
+def register_mcp_tool(mcp_func: Callable, mcp: FastMCP) -> Callable:
```
NISH1001: change the type hint to be precise:

`Callable[..., Awaitable[Any]]`

akd_ext/mcp/mcp/registry.py
```diff
@@ -0,0 +1,72 @@
...
+        self._tools.clear()
+
+    @classmethod
+    def _reset_singleton(cls) -> None:
+        """Reset the singleton instance (for testing purposes only)."""
+        cls._instance = None
+        cls._initialized = False
+
+
+mcp_tool_registry = MCPToolRegistry()
```
NISH1001: Capitalize global variable.

`MCP_TOOL_REGISTRY`.

Better, we can just do this initialization in `server.py` since it's a singleton. so other places we can just get object/initialize and should give same object.

akd_ext/mcp/mcp/server.py
```diff
@@ -0,0 +1,58 @@
...
+    """
+    Auto-discover and register all @mcp_tool decorated classes.
+
+    This function imports the tools module to trigger decorator registration,
+    then converts and registers each tool with the FastMCP server.
+
+    Example:
+        register_all_tools() 
+    """
+    from akd_ext import tools
```
NISH1001: import at module-level if possible

akd_ext/mcp/mcp/registry.py
NISH1001: Alterantely: have a proxy function that does `def get_mcp_tool_registry(...)`.

```python
def get_mcp_tool_registry(...):
  return MCPToolRegistry(...)
```

I like this because this avoids having conflict that "This is an instance that should be used". Instead, we can just get the instance wherever we want by doing `get_mcp_tool_registry`.

or even wherever we can just import and use


```python
from akd_ext.mcp.registry import MCPToolRegistry

...

# say random_file.py

def some_fn(tool):
...
...
registry = MCPToolRegistry(...)
regsitry.register(tool)
```

I like this approach because we already have established the registry is singleton, so it's guaranteed to have only single instance reused. and avoids us defining global variables...

akd_ext/mcp/mcp/server.py
```diff
@@ -0,0 +1,58 @@
...
+
+    for tool_class in tool_classes:
+        tool = tool_class()
+        # Convert tool to FastMCP compatible function
+        mcp_func = tool_converter(tool)
+        # Register tool with FastMCP server
+        register_mcp_tool(mcp_func, mcp)
+
+
+def register_manual_tools():
```
NISH1001: can you have this take tools as list input

```python
def register_tools_manually(tools: list[BaseTool])...
...
for tool_cls in tools:
..
..
```

This way you can just have 

```python
# top-level

register_all_available_tools(...)

regsitry_tools_manually(tools=[
...
...
])
```

NISH1001: rename to `register_tools_manually`

pyproject.toml
```diff
@@ -17,7 +17,9 @@ classifiers = [
 requires-python = ">=3.12"
 dependencies = [
     "akd @ git+https://github.com/NASA-IMPACT/accelerated-discovery.git@develop",
+    "fastmcp>=2.0.0",
     "openai-agents>=0.6.7",
+    "pydantic>=2.0.0",
```
NISH1001: i think this pydantic deps will already be available via akd core.

NISH1001 (changes requested):

akd_ext/mcp/mcp/converter.py
rohit-sahoo: Updated!

akd_ext/mcp/mcp/converter.py
rohit-sahoo: Updated!

akd_ext/mcp/mcp/registry.py
rohit-sahoo: Updated! using MCPToolRegistry() directly, making sure singleton makes same instance everywhere. No global variables needed.

akd_ext/mcp/mcp/server.py
rohit-sahoo: Updated!

akd_ext/mcp/mcp/server.py
rohit-sahoo: Updated!, taking tools as list parameter

rohit-sahoo: updated!

pyproject.toml
rohit-sahoo: removed!

NISH1001:
@rohit-sahoo thanks for the quick fixes. I will pull this to a private fork and test and confirm if it works. Will be merged by first half tomorrow. THanks.

NISH1001:
@rohit-sahoo one last ask: can you also add logger.info stuff to server.py or when registering mcp so that we know what has been registered....That will give us idea when server starts we get some visible signals

NISH1001:
Also getting this error in python 3.12:

>                                                                                                                                                                                               
⏺ Bash(uv run python akd_ext/mcp/server.py) timeout: 30s                                                                                                                                      
  ⎿  Error: Exit code 1                                                                                                                                                                       
     Traceback (most recent call last):                                                                                                                                                       
       File "/Users/npantha/dev/impact/projects/nish1001-akd-ext/akd_ext/mcp/server.py", line 3, in <module>                                                                                  
         from fastmcp import FastMCP                                                                                                                                                          
       File "/Users/npantha/dev/impact/projects/nish1001-akd-ext/.venv/lib/python3.12/site-packages/fastmcp/__init__.py", line 15, in <module>                                                
         from fastmcp.server.server import FastMCP                                                                                                                                            
       File "/Users/npantha/dev/impact/projects/nish1001-akd-ext/.venv/lib/python3.12/site-packages/fastmcp/server/__init__.py", line 1, in <module>                                          
         from .server import FastMCP                                                                                                                                                          
       File "/Users/npantha/dev/impact/projects/nish1001-akd-ext/.venv/lib/python3.12/site-packages/fastmcp/server/server.py", line 34, in <module>                                           

Claude saying:                                                                                                                                                                                              
⏺ There's a dependency issue. The wrapt package version is incompatible with Python 3.12 (formatargspec was removed from inspect in Python 3.11).                                             
                             

Here's what I get:

```
npantha@nishan24-mbp nish1001-akd-ext % uv run fastmcp run akd_ext/mcp/server.py:mcp
Traceback (most recent call last):
  File "/Users/npantha/dev/impact/projects/nish1001-akd-ext/.venv/bin/fastmcp", line 4, in <module>
    from fastmcp.cli import app
  File "/Users/npantha/dev/impact/projects/nish1001-akd-ext/.venv/lib/python3.12/site-packages/fastmcp/__init__.py", line 15, in <module>
    from fastmcp.server.server import FastMCP
  File "/Users/npantha/dev/impact/projects/nish1001-akd-ext/.venv/lib/python3.12/site-packages/fastmcp/server/__init__.py", line 1, in <module>
    from .server import FastMCP
  File "/Users/npantha/dev/impact/projects/nish1001-akd-ext/.venv/lib/python3.12/site-packages/fastmcp/server/server.py", line 34, in <module>
    from docket import Docket, Worker
  File "/Users/npantha/dev/impact/projects/nish1001-akd-ext/.venv/lib/python3.12/site-packages/docket/__init__.py", line 11, in <module>
    from .agenda import Agenda
  File "/Users/npantha/dev/impact/projects/nish1001-akd-ext/.venv/lib/python3.12/site-packages/docket/agenda.py", line 14, in <module>
    from .docket import Docket
  File "/Users/npantha/dev/impact/projects/nish1001-akd-ext/.venv/lib/python3.12/site-packages/docket/docket.py", line 40, in <module>
    from .strikelist import (
  File "/Users/npantha/dev/impact/projects/nish1001-akd-ext/.venv/lib/python3.12/site-packages/docket/strikelist.py", line 28, in <module>
    from opentelemetry.instrumentation.utils import suppress_instrumentation
  File "/Users/npantha/dev/impact/projects/nish1001-akd-ext/.venv/lib/python3.12/site-packages/opentelemetry/instrumentation/utils.py", line 23, in <module>
    from wrapt import ObjectProxy
  File "/Users/npantha/dev/impact/projects/nish1001-akd-ext/.venv/lib/python3.12/site-packages/wrapt/__init__.py", line 10, in <module>
    from .decorators import (adapter_factory, AdapterFactory, decorator,
  File "/Users/npantha/dev/impact/projects/nish1001-akd-ext/.venv/lib/python3.12/site-packages/wrapt/decorators.py", line 34, in <module>
    from inspect import ismethod, isclass, formatargspec
ImportError: cannot import name 'formatargspec' from 'inspect' (/Users/npantha/.local/share/uv/python/cpython-3.12.0-macos-aarch64-none/lib/python3.12/inspect.py). Did you mean: 'formatargvalues'?

```

NISH1001:
I think there's some issue with these dependency.

After going back and forth with claude:
```
  Quickest fix - add override to pyproject.toml:                                                                                                                                              
                                                                                                                                                                                              
  [tool.uv]                                                                                                                                                                                
  override-dependencies = ["wrapt>=1.14"]    
```

NISH1001:
Also it seems there's circular import issue when we apply the `mcp_tool` decorator to the tools.

```
server.py → tools → dummy.py → mcp → server.py → dummy.py 
```

We should fix this

NISH1001:
I think you might have to use lazy import in `server.py`. I will let you test and fix these.

RUn it locally `uv run fastmcp run akd_ext/mcp/server.py:mcp`

akd_ext/mcp/server.py
```diff
@@ -0,0 +1,55 @@
...
+        tool = tool_class()
+        mcp_func = tool_converter(tool)
+        register_mcp_tool(mcp_func, mcp)
+
+
+register_all_tools()
+# register_tools_manually(tools=[])  # Add tools here if needed
+
+if __name__ == "__main__":
+    mcp.run()
```
NISH1001: we need to account for some paramters to the run function as well. Eg: host and port, both get from `os.getenv("MCP_HOST", "0.0.0.0")` and `os.getenv("MCP_PORT", `)?

Or better use argparse to control?

```
 57  if __name__ == "__main__":                                                                                                                                                                                                                                                                                                                               
 58 +    import argparse                                                                                                                                                                      
 59 +    import os                                                                                                                                                                            
 60 +                                                                                                                                                                                         
 61 +    parser = argparse.ArgumentParser(description="Run akd-ext MCP server")                                                                                                               
 62 +    parser.add_argument(                                                                                                                                                                 
 63 +        "--transport",                                                                                                                                                                   
 64 +        choices=["stdio", "sse"],                                                                                                                                                        
 65 +        default=os.getenv("MCP_TRANSPORT", "stdio"),                                                                                                                                     
 66 +        help="Transport type (default: stdio, or MCP_TRANSPORT env var)",                                                                                                                
 67 +    )                                                                                                                                                                                    
 68 +    parser.add_argument(                                                                                                                                                                 
 69 +        "--host",                                                                                                                                                                        
 70 +        default=os.getenv("MCP_HOST", "127.0.0.1"),                                                                                                                                      
 71 +        help="Host to bind to for SSE (default: 127.0.0.1)",                                                                                                                             
 72 +    )                                                                                                                                                                                    
 73 +    parser.add_argument(                                                                                                                                                                 
 74 +        "--port",                                                                                                                                                                        
 75 +        type=int,                                                                                                                                                                        
 76 +        default=int(os.getenv("MCP_PORT", "8000")),                                                                                                                                      
 77 +        help="Port for SSE transport (default: 8000)",                                                                                                                                   
 78 +    )                                                                                                                                                                                    
 79 +    args = parser.parse_args()                                                                                                                                                           
 80 +                                                                                                                                                                                         
 81 +    if args.transport == "sse":                                                                                                                                                          
 82 +        mcp.run(transport="sse", host=args.host, port=args.port)                                                                                                                         
 83 +    else:                                                                                                                                                                                
 84 +        mcp.run()
```

NISH1001: This to make sure we can do local testing accordingly? or spawn the server if hosted manually? I odn't think this is needed for fastmcpcloud. but good to have for future if we need to host somewhere with more control?

NISH1001: Okay. I think #14 fixes this

NISH1001:
@rohit-sahoo Okay. I think #14 fixes this. can you look at that and test?

NISH1001 (approved):

PR #19, opened 2026-01-26
set retry to false. also add in try catch to handle extra issues

sanzog03:
# Issue:
Exponential backoff policy of PyGithub causing MCP cloud to fail.

# resolution:
The rate limit is minimal for github without access token. So, instead of waiting for the response via exponential backoff, cutting off the request and sending back default values.

To elobrate, the PyGithub has an elaborate retry mechanism and timeouts built out of the box. As the default timeout is 15 seconds, i just left that as it is. I've just changed the retry to be none.

The alternative method would be to not have a concesion on DEFAULT_BACKOFF_MAX (120 current) and DEFAULT_RETRY_AFTER_MAX(21600 current) to something smaller.

### DIFF
--- a/akd_ext/tools/code_search/utils.py
+++ b/akd_ext/tools/code_search/utils.py
@@ -52,7 +52,7 @@ async def fetch_github_metadata(repo_name: str, access_token: str | None = None)
     return repository_metadata
 
 
-def calculate_reliability_score(repository_metadata: RepositoryMetadata) -> float:
+def calculate_reliability_score(repository_metadata: RepositoryMetadata) -> float | None:
     """
     Calculate a comprehensive reliability score (0-100) for a repository.
 
@@ -71,6 +71,10 @@ def calculate_reliability_score(repository_metadata: RepositoryMetadata) -> floa
     Returns:
         float: reliability score between 0 and 100
     """
+    # if repository_metadata is default return None
+    if repository_metadata == RepositoryMetadata():
+        return None
+
     now = datetime.now(timezone.utc)
 
     # Parse dates - return 0 if essential data is missing
--- a/akd_ext/tools/code_search/utils.py
+++ b/akd_ext/tools/code_search/utils.py
@@ -33,17 +33,22 @@ async def fetch_github_metadata(repo_name: str, access_token: str | None = None)
     auth = None
     if access_token:
         auth = Auth.Token(access_token)
-    with Github(auth=auth) as g:
-        repo = g.get_repo(repo_name)
-        repository_metadata.stars = repo.stargazers_count
-        repository_metadata.forks = repo.forks_count
-        repository_metadata.watchers = repo.subscribers_count
-        repository_metadata.last_updated = repo.pushed_at.isoformat() if repo.pushed_at else ""
-        repository_metadata.created_at = repo.created_at.isoformat() if repo.created_at else ""
-        repository_metadata.open_issues = repo.get_issues(state="open").totalCount
-        repository_metadata.pulls = repo.get_pulls(state="open", sort="created", base="master").totalCount
-        repository_metadata.closed_pulls = repo.get_pulls(state="closed", sort="created", base="master").totalCount
-        repository_metadata.first_commit_date = None  # The original comde provided also fell back to created_at if first_commit_date was not available. And it was set to None by default.
+    try:
+        with Github(auth=auth, retry=None) as g:
+            repo = g.get_repo(repo_name)
+            repository_metadata.stars = repo.stargazers_count
+            repository_metadata.forks = repo.forks_count
+            repository_metadata.watchers = repo.subscribers_count
+            repository_metadata.last_updated = repo.pushed_at.isoformat() if repo.pushed_at else ""
+            repository_metadata.created_at = repo.created_at.isoformat() if repo.created_at else ""
+            repository_metadata.open_issues = repo.get_issues(state="open").totalCount
+            repository_metadata.pulls = repo.get_pulls(state="open", sort="created", base="master").totalCount
+            repository_metadata.closed_pulls = repo.get_pulls(state="closed", sort="created", base="master").totalCount
+            repository_metadata.first_commit_date = None  # The original code provided also fell back to created_at if first_commit_date was not available. And it was set to None by default.
+    except Exception as e:
+        logger.error(f"Error fetching metadata for {repo_name}: {e}")
+        return repository_metadata
+
     return repository_metadata
 
 
--- a/akd_ext/tools/code_search/utils.py
+++ b/akd_ext/tools/code_search/utils.py
@@ -1,5 +1,5 @@
 from github import Github, Auth
-from pydantic import BaseModel, Field
+from pydantic import BaseModel, Field, computed_field
 from datetime import datetime, timezone
 import math
 from functools import lru_cache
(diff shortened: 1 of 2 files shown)


NISH1001:
@sanzog03 also add asyncio.timeout() iwthin the tool when you fetch the metadata to make sure it doesn't hold longer time.

```python
import asyncio

class Tool(...):

Config: github_timeout: int = FIeld(default=int(os.getenv("GITHUB_API_TIMEOUT", "30")))
...
def _arun(...):...
...
...
repo_list = <some_list>

# maybe put asyncio.timeout() to each task
metadata_list = asyncio.gather(*tasks)
```

Handling exeptoin is one way. But gracefully doing timeout is another way. I will let you handle. Maybe even in the function you fetch we can add timeout.

sanzog03:
> @sanzog03 also add asyncio.timeout() iwthin the tool when you fetch the metadata to make sure it doesn't hold longer time.
> 
> ```python
> import asyncio
> 
> class Tool(...):
> 
> Config: github_timeout: int = FIeld(default=int(os.getenv("GITHUB_API_TIMEOUT", "30")))
> ...
> def _arun(...):...
> ...
> ...
> repo_list = <some_list>
> 
> # maybe put asyncio.timeout() to each task
> metadata_list = asyncio.gather(*tasks)
> ```
> 
> Handling exeptoin is one way. But gracefully doing timeout is another way. I will let you handle. Maybe even in the function you fetch we can add timeout.

updated the description. The default timeout is 15 secs. so adding it explicitly would be redundant

NISH1001:
We also need to make sure we're not retrying anything because the main goal is to fetch repo list. So, we don't have to fetch metadata if it causing issue once. I am afraid if the module `github` or something would internally do some backoff. I am sure it's doing some backoff. So check the parameter there as well.

NISH1001:
> > @sanzog03 also add asyncio.timeout() iwthin the tool when you fetch the metadata to make sure it doesn't hold longer time.
> > ```python
> > import asyncio
> > 
> > class Tool(...):
> > 
> > Config: github_timeout: int = FIeld(default=int(os.getenv("GITHUB_API_TIMEOUT", "30")))
> > ...
> > def _arun(...):...
> > ...
> > ...
> > repo_list = <some_list>
> > 
> > # maybe put asyncio.timeout() to each task
> > metadata_list = asyncio.gather(*tasks)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Handling exeptoin is one way. But gracefully doing timeout is another way. I will let you handle. Maybe even in the function you fetch we can add timeout.
> 
> updated the description. The default timeout is 15 secs. so adding it explicitly would be redundant

Oh okay. Didn't know. Then that's good to see

sanzog03:
> As the default timeout is 15 seconds, i just left that as it is.

As the default timeout is 15 seconds, i just left that as it is. Should the timeout be smaller than that?

akd_ext/tools/code_search/utils.py
```diff
@@ -33,17 +33,22 @@ async def fetch_github_metadata(repo_name: str, access_token: str | None = None)
...
+            repository_metadata.watchers = repo.subscribers_count
+            repository_metadata.last_updated = repo.pushed_at.isoformat() if repo.pushed_at else ""
+            repository_metadata.created_at = repo.created_at.isoformat() if repo.created_at else ""
+            repository_metadata.open_issues = repo.get_issues(state="open").totalCount
+            repository_metadata.pulls = repo.get_pulls(state="open", sort="created", base="master").totalCount
+            repository_metadata.closed_pulls = repo.get_pulls(state="closed", sort="created", base="master").totalCount
+            repository_metadata.first_commit_date = None  # The original code provided also fell back to created_at if first_commit_date was not available. And it was set to None by default.
+    except Exception as e:
+        logger.error(f"Error fetching metadata for {repo_name}: {e}")
+        return repository_metadata
```
NISH1001, line 58: by defualt everything is None in the metadata, right?

NISH1001:
> > As the default timeout is 15 seconds, i just left that as it is.
> 
> As the default timeout is 15 seconds, i just left that as it is. Should the timeout be smaller than that?

It's fine for now.

sanzog03:
with retry set to default:
image

sanzog03:
with retry set to none

image

akd_ext/tools/code_search/utils.py
sanzog03: its basically zero or alternative basde on the data type.

sanzog03: maybe the screenshot gives more context.

NISH1001:
@sanzog03 can we set reliability score to None or something when we hit github error or something? so that, we won't let agent misinterpret `0` value. None could be undefined, not computed. 0 means numerical 0 which agent can misinterpret as "most unreliable". Let's fix that as well. Default the `reliabilitY_score` to none and set only when metadata is available. If everything is NOne/0, then no need to set.

NISH1001:
Oh I htink default 1.0 is there. SHould we set to None or 1.0? I guess default to 1.0 is okay and sets everything as reliable. which is good. Just do the validation that if there's error, we aren't setting or computing the score to 0.0 to prevent any misinterpretation by the agent.

sanzog03:
I like None better, to make it netural rather than giving out 1 (max) reliability. Say, there are 10 repos and 8 could not fetch metadata, this could make the relevant 2 not appear in the result as the other 8 will have hightest reliability score.

sanzog03:
image

akd_ext/tools/code_search/utils.py
```diff
@@ -66,6 +71,10 @@ def calculate_reliability_score(repository_metadata: RepositoryMetadata) -> floa
     Returns:
         float: reliability score between 0 and 100
     """
+    # if repository_metadata is default return None
+    if repository_metadata == RepositoryMetadata():
+        return None
```
NISH1001: it should happen at field-level all() check. this is not the correct way to check unless there's `__eq__` overridden in pydantic model. Or if pydnatic already behaves that way.

NISH1001 (changes requested):

akd_ext/tools/code_search/utils.py
NISH1001: Also, `reliability_score` field in the basemodel would have float | None type and Field(default=None,...) would also be something.

NISH1001:
```python
class RepositoryMetadata(BaseModel):
    """
    Github repository metadata. includes number of stars, forks, open issues, open pull requests, and closed pull requests.
    """

    # reliability metrics
    stars: int = Field(default=0, description="Number of stars on the repository.")
    forks: int = Field(default=0, description="Number of forks on the repository.")
    watchers: int = Field(default=0, description="Number of watchers on the repository.")
    last_updated: str = Field(default="", description="Last updated date of the repository.")
    created_at: str = Field(default="", description="Creation date of the repository.")
    first_commit_date: str = Field(default="", description="Date of the first commit in the repository.")
    # few extra metrics
    open_issues: int = Field(default=0, description="Number of open issues on the repository.")
    pulls: int = Field(default=0, description="Number of open pull requests on the repository.")
    closed_pulls: int = Field(default=0, description="Number of closed pull requests on the repository.")

   @computed_field
  def is_null_metadata(self) -> bool:
    return self.stars ==0 and self.forks=0, ......
```

here, either implement a @computed_field that says is_null_metadata() where you check stars , forks etc all 0. then you'll know it's an empty metadata.

NISH1001: or just put RepositoryMetadata as optoinal None in the output schema. that should fix it.

NISH1001: Pydantic does have default `__eq__` implementedd, but the object equality check is a bit fragile with this because we have so many things. I'd rather do a computed field saying is_null() where we check stars is 0, forks is 0, watchers i0, last_update is empty. etc....

NISH1001: Or just make metadata optional in the output schema

sanzog03: I like the idea of adding is_null_metadata as a computed field property under RepositoryMetadata. I think it will providing additional context about reliability score calculation.
Just made the changes

NISH1001 (approved):

PR #21, opened 2026-01-27
Fix/codesearch optimizations and tests

sanzog03:
## Issue
1. `@lru_cache` decorator on async functions causing `RuntimeError: cannot reuse already awaited coroutine` when cached coroutines are awaited multiple times
2. Missing null safety check in [calculate_reliability_score](cci:1://file:///Users/thapa24-mbp/Devs/akd/akd-ext/akd_ext/tools/code_search/utils.py:62:0-136:32) causing potential `AttributeError` 
3. Insufficient test coverage for code search utilities

## Resolution
1. **Fixed async caching mechanism**: Replaced `@lru_cache` with `@async_lru_cache` from the `akd.utils.async_lru_cache` package for the [fetch_github_metadata](cci:1://file:///Users/thapa24-mbp/Devs/akd/akd-ext/akd_ext/tools/code_search/utils.py:34:0-59:30) function. The `@lru_cache` caches the coroutine object itself rather than the result, while `@alru_cache` properly caches the awaited results of async functions, allowing concurrent requests with identical arguments.
2. **Improved reliability score calculation**: Added null safety check to handle `None` input in [calculate_reliability_score](cci:1://file:///Users/thapa24-mbp/Devs/akd/akd-ext/akd_ext/tools/code_search/utils.py:62:0-136:32):
   ```python
   if not repository_metadata or repository_metadata.is_null_metadata:
       return None
       
3. **Added comprehensive test coverage**:
    - Created tests/tools/code_search/test_utils.py with tests for successful metadata fetching, failure handling for non-existent repositories, reliability score calculations, and null metadata edge cases
    - Enhanced tests/tools/code_search/test_repository_search.py with parametrized tests for page size variations (3, 8, 10, 11, 55, 100) to verify results are capped at maximum 10 regardless of page_size setting

### DIFF
--- a/akd_ext/tools/code_search/utils.py
+++ b/akd_ext/tools/code_search/utils.py
@@ -80,7 +80,7 @@ def calculate_reliability_score(repository_metadata: RepositoryMetadata) -> floa
         float: reliability score between 0 and 100 or None if metadata is null
     """
 
-    if repository_metadata.is_null_metadata:
+    if not repository_metadata or repository_metadata.is_null_metadata:
         return None
 
     now = datetime.now(timezone.utc)
--- a/akd_ext/tools/code_search/utils.py
+++ b/akd_ext/tools/code_search/utils.py
@@ -2,7 +2,7 @@ from github import Github, Auth
 from pydantic import BaseModel, Field, computed_field
 from datetime import datetime, timezone
 import math
-from functools import lru_cache
+from async_lru import alru_cache
 from loguru import logger
 
 
@@ -32,7 +32,7 @@ class RepositoryMetadata(BaseModel):
         )
 
 
-@lru_cache(maxsize=128)
+@alru_cache(maxsize=128)
 async def fetch_github_metadata(repo_name: str, access_token: str | None = None) -> RepositoryMetadata:
     """
     Repo_name should be in the format of owner/repo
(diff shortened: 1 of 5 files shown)


NISH1001:
@sanzog03 we have https://github.com/NASA-IMPACT/accelerated-discovery/blob/develop/akd/utils.py#L19
this already in akd core async lru cache. can you try that since it's arlady in akd core so.

akd_ext/tools/code_search/utils.py
```diff
@@ -32,7 +32,7 @@ def is_null_metadata(self) -> bool:
         )
 
 
-@lru_cache(maxsize=128)
+@alru_cache(maxsize=128)
```
NISH1001: we also have `akd.utils.async_lru_cache` which could be self-contained? no external deps

sanzog03: changed!

sanzog03:
> @sanzog03 we have https://github.com/NASA-IMPACT/accelerated-discovery/blob/develop/akd/utils.py#L19 this already in akd core async lru cache. can you try that since it's arlady in akd core so.

changed!

uv.lock
NISH1001, line 1: Is this the updated uv lock? What changes occur here? did you add new pcakagges?

NISH1001: I just want to make sure we're not pushing unnecessary local package changes that are not required for akd-ext functioning

sanzog03: yes. updated the uv lock after removing the unwanted package.

NISH1001 (approved):

PR #27, opened 2026-02-02
MCP compatible STAC collections RAG search tool for EIE agent

lillythomas:
## Summary

Add a MCP-compatible tool for semantic search over STAC collections using LanceDB and Ollama embeddings, enabling RAG-powered dataset discovery for the EIE agent.

## What it does

- Performs vector similarity search over STAC collection metadata using natural language queries (e.g., "NO2 air quality", "sea surface temperature")
- Checks spatial and temporal overlap between user's requested extent and each collection's coverage
- Returns matched collection IDs with `spatial_overlap` and `temporal_overlap` flags to help filter relevant datasets

## How it does

1. **Added CollectionsRagTool class** inheriting from `BaseTool[CollectionsRagInputSchema, CollectionsRagOutputSchema]`
2. **Implemented async _arun() method**
   - Generates query embeddings via Ollama API (`nomic-embed-text` model)
   - Searches LanceDB vector index using cosine similarity
   - Computes spatial overlap between user bbox and collection extents
   - Computes temporal overlap between user datetime range and collection intervals
3. **Added configuration via CollectionsRagToolConfig**
   - `db_path`: Path to LanceDB database (from `COLLECTIONS_RAG_DB_PATH` env var)
   - `ollama_url`: Ollama API URL (from `OLLAMA_URL` env var)
   - `embedding_model`: Model name (default: `nomic-embed-text`)
   - `timeout`: HTTP timeout (default: 60s)
4. **Added helper functions** for bbox and temporal interval overlap checking

## Files changed

- akd_ext/tools/collections_rag.py — New tool implementation
- akd_ext/tools/__init__.py — Added exports
- pyproject.toml — Added `lancedb` dependency

## Testing

```python
import asyncio
from akd_ext.tools import CollectionsRagTool
from akd_ext.tools.collections_rag import CollectionsRagInputSchema

async def test():
    tool = CollectionsRagTool()
    result = await tool.arun(CollectionsRagInputSchema(
        query="NO2 air quality",
        bbox=[-124.41, 32.53, -114.13, 42.01],
        datetime="2021-10-01/2021-12-31"
    ))
    print("Collections:", result.collections)
    for m in result.matches:
        print(f"  - {m.id}: spatial={m.spatial_overlap}, temporal={m.temporal_overlap}")

asyncio.run(test())
```

**Output:**
```
Collections: ['omi-no2-2d', 'no2-monthly-diff', 'no2-monthly']
  - omi-no2-2d: spatial=True, temporal=False
  - no2-monthly-diff: spatial=True, temporal=True
  - no2-monthly: spatial=True, temporal=True
```

**Environment variables required:**
- `COLLECTIONS_RAG_DB_PATH` — Path to LanceDB with collection embeddings
- `OLLAMA_URL` — Ollama embeddings API URL (default: `http://localhost:11434`)

**Note:** The LanceDB index must be pre-built with collection embeddings before using this tool.

### DIFF
--- a/akd_ext/tools/collections_rag.py
+++ b/akd_ext/tools/collections_rag.py
@@ -1,0 +1,296 @@
+"""
+STAC Collections RAG tool for semantic search over Earth science data collections.
+
+This tool uses vector embeddings (via Ollama) and LanceDB to perform semantic
+search over STAC collection metadata, helping users find relevant Earth science
+datasets based on natural language queries.
+"""
+
+import os
+from datetime import datetime, timezone
+from typing import Literal
+
+import httpx
+import lancedb
+from akd._base import InputSchema, OutputSchema
+from akd.tools import BaseTool, BaseToolConfig
+from pydantic import Field
+from loguru import logger
+
+from akd_ext.mcp import mcp_tool
+
+
+class CollectionsRagToolConfig(BaseToolConfig):
+    """Configuration for the Collections RAG Tool."""
+
+    db_path: str = Field(
+        default=os.getenv("COLLECTIONS_RAG_DB_PATH", "/tmp/veda_collections.lancedb"),
+        description="Path to the LanceDB database containing collection embeddings",
+    )
+    ollama_url: str = Field(
+        default=os.getenv("OLLAMA_URL", "http://localhost:11434"),
+        description="Base URL for the Ollama embeddings API",
+    )
+    embedding_model: str = Field(
+        default=os.getenv("COLLECTIONS_RAG_MODEL", "nomic-embed-text"),
+        description="Name of the Ollama embedding model to use",
+    )
+    timeout: float = Field(
+        default=60.0,
+        description="HTTP request timeout for embedding requests in seconds",
+    )
+    top_k: int = Field(
+        default=3,
+        ge=1,
+        le=20,
+        description="Default number of results to return",
+    )
+
+
+class CollectionMatchInfo(OutputSchema):
+    """Information about a matched collection including coverage overlap."""
+
+    id: str = Field(..., description="STAC collection ID")
+    title: str | None = Field(None, description="Collection title")
+    spatial_overlap: bool = Field(
+        default=True,
+        description="Whether the collection spatially overlaps the requested bbox",
+    )
+    temporal_overlap: bool = Field(
+        default=True,
+        description="Whether the collection temporally overlaps the requested time range",
+    )
+
+
+class CollectionsRagInputSchema(InputSchema):
+    """Input schema for the Collections RAG tool."""
+
+    query: str = Field(
+        ...,
+        description="Natural language description of the data you're looking for (e.g., 'NO2 air quality', 'sea surface temperature')",
+    )
+    bbox: list[float] | None = Field(
+        None,
+        description="Optional bounding box [west, south, east, north] to check spatial coverage",
+    )
+    datetime: str | None = Field(
+        None,
+        description="Optional ISO-8601 datetime range (e.g., '2021-10-01/2021-12-31') to check temporal coverage",
+    )
+    limit: int = Field(
+        default=3,
+        ge=1,
+        le=20,
+        description="Maximum number of collections to return",
+    )
+
+
+class CollectionsRagOutputSchema(OutputSchema):
+    """Output schema for the Collections RAG tool."""
+
+    collections: list[str] = Field(
+        default_factory=list,
+        description="List of matched collection IDs",
+    )
+    matches: list[CollectionMatchInfo] = Field(
+        default_factory=list,
+        description="Detailed match information including spatial/temporal overlap flags",
+    )
+    error: str | None = Field(
+        None,
+        description="Error message if search failed",
+    )
+
+
+def _bboxes_overlap(bbox1: list[float], bbox2: list[float]) -> bool:
+    """Check if two bounding boxes overlap.
+    
+    Args:
+        bbox1: First bounding box as [west, south, east, north]
+        bbox2: Second bounding box in the same format
+    
+    Returns:
+        True if the boxes overlap, False otherwise
+    """
+    if len(bbox1) < 4 or len(bbox2) < 4:
+        return False
+    
+    w1, s1, e1, n1 = bbox1[:4]
+    w2, s2, e2, n2 = bbox2[:4]
+    
+    # Two boxes do NOT overlap if one is entirely to the left, right, above, or below
+    return not (e1 < w2 or e2 < w1 or n1 < s2 or n2 < s1)
+
+
+def _parse_iso_date(s: str | None) -> datetime | None:
+    """Parse ISO-8601 date string to UTC-aware datetime."""
+    if not s:
+        return None
+    try:
+        dt = datetime.fromisoformat(s.replace("Z", "+00:00"))
+        if dt.tzinfo is None:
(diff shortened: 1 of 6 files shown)


akd_ext/tools/collections_rag.py
```diff
@@ -0,0 +1,296 @@
...
+    async def _arun(self, params: CollectionsRagInputSchema) -> CollectionsRagOutputSchema:
+        """Execute semantic search over collections and return matches."""
+        try:
+            # Generate query embedding
+            logger.debug(f"Generating embedding for query: {params.query[:100]}")
+            query_vector = await self._embed_query(params.query)
+
+            # Connect to LanceDB and search
+            db = lancedb.connect(self.config.db_path)
+            table = db.open_table("veda_collections")
```
NISH1001: maybe have this colleciton name in the config as well and default to veda_collections.

akd_ext/tools/eie/collections_rag.py
```diff
@@ -0,0 +1,296 @@
...
+search over STAC collection metadata, helping users find relevant Earth science
+datasets based on natural language queries.
+"""
+
+import os
+from datetime import datetime, timezone
+from typing import Literal
+
+import httpx
+import lancedb
```
NISH1001, line 13: Any reason we choose lancedb over other vectordb like qdrant?

akd_ext/tools/eie/collections_rag.py
NISH1001, line 1: Better to have the namespace to something like `akd_ext.tools.veda/collections_rag.py`  that way we can have proper namespace to import from.

akd_ext/tools/eie/collections_rag.py
```diff
@@ -0,0 +1,296 @@
...
+            )
+            response.raise_for_status()
+            embedding = response.json().get("embedding")
+            if not embedding:
+                raise ValueError(f"No embedding returned from Ollama for query: {query[:100]}")
+            return [float(x) for x in embedding]
+
+    async def _arun(self, params: CollectionsRagInputSchema) -> CollectionsRagOutputSchema:
+        """Execute semantic search over collections and return matches."""
+        try:
```
NISH1001, line 163: narrow down the try catch to specifc set of blocks that is likely to lead to error instead of having whole code wrapped in a big try-catch. It will improve debugging.

akd_ext/tools/collections_rag.py
```diff
@@ -0,0 +1,296 @@
...
+                        temporal_overlap=temporal_overlap,
+                    )
+                )
+
+            return CollectionsRagOutputSchema(
+                collections=collection_ids,
+                matches=matches,
+                error=None,
+            )
+
```
NISH1001: I'd try to avoid this early return. Maybe change the flow so that return will be done at the end of the method rather than in the middle right before try catch. 

```python
def _arun(...):

final_output = CollectionsRagOutputSchema() # empty 

do things
do things again
try:
...
...
final_output = ...
except:
...

return final_output
```

This makes it more "flowy".

akd_ext/tools/collections_rag.py
```diff
@@ -0,0 +1,296 @@
...
+                .limit(params.limit)
+                .to_list()
+            )
+
+            # Parse datetime range if provided
+            req_start, req_end = None, None
+            if params.datetime and "/" in params.datetime:
+                parts = params.datetime.split("/")
+                if len(parts) == 2:
+                    req_start, req_end = parts[0].strip(), parts[1].strip()
```
NISH1001: What is the expected format? Maybe put in the example so that in future anyone who looks at code would know the expected format and know why this parsing is done.

akd_ext/tools/collections_rag.py
```diff
@@ -0,0 +1,296 @@
...
+    
+    # Collection ends before requested range starts → no overlap
+    if col_end and req_start and col_end < req_start:
+        return False
+    
+    # Collection starts after requested range ends → no overlap
+    if col_start and req_end and col_start > req_end:
+        return False
+    
+    return True
```
NISH1001: Assuming we have moved to new namespace like `akd_ext.tools.veda`, we can have these as reusable utils within `akd_ext.tools.veda.utils.py`.

NISH1001 (changes requested):

akd_ext/tools/eie/collections_rag.py
lillythomas: Addressed in https://github.com/NASA-IMPACT/akd-ext/pull/27/commits/c3bb3d72ef8098e65ac5708b1b6bfb7af50fc6d5. Note I used `eie` as the namespace as we will likely involve more collection sources than veda in the future.

akd_ext/tools/collections_rag.py
lillythomas: Addressed in https://github.com/NASA-IMPACT/akd-ext/pull/27/commits/c3bb3d72ef8098e65ac5708b1b6bfb7af50fc6d5.

akd_ext/tools/eie/collections_rag.py
lillythomas: Addressed in https://github.com/NASA-IMPACT/akd-ext/pull/27/commits/c3bb3d72ef8098e65ac5708b1b6bfb7af50fc6d5.

akd_ext/tools/collections_rag.py
lillythomas: Addressed in https://github.com/NASA-IMPACT/akd-ext/pull/27/commits/c3bb3d72ef8098e65ac5708b1b6bfb7af50fc6d5.

akd_ext/tools/collections_rag.py
lillythomas: Addressed in https://github.com/NASA-IMPACT/akd-ext/pull/27/commits/c3bb3d72ef8098e65ac5708b1b6bfb7af50fc6d5. 

See: https://github.com/NASA-IMPACT/akd-ext/pull/27/changes/c3bb3d72ef8098e65ac5708b1b6bfb7af50fc6d5#diff-b6a78bb6785b98f50f9e0da4da7f817502c1961a9c092cb770db593e088acb03R205-R206

akd_ext/tools/collections_rag.py
lillythomas: Addressed in [c3bb3d7](https://github.com/NASA-IMPACT/akd-ext/pull/27/commits/c3bb3d72ef8098e65ac5708b1b6bfb7af50fc6d5).

akd_ext/tools/eie/collections_rag.py
lillythomas: For our current use case (small, static collection metadata of a few hundred vectors), LanceDB is a pragmatic choice. If we need to scale to support more, we can look at something like qdrant. Open to this discussion on alternative choices though.

lillythomas:
Thanks @NISH1001 for your review. I've addressed your current comments. Please let me know if you have further feedback and suggested changes.

akd_ext/tools/eie/collections_rag.py
NISH1001: Eh okay. For now i think lancedb is good. i think we need to have a spearate discussion for vectorstore in akd itself in separate thread. We've been thinking about this.

NISH1001:
@lillythomas thanks for the changes. I resolved the comments. Looks good. I will do another pass at it in detail if I missed anything. And then let you know. Thanks.

PR #28, opened 2026-02-03
feat: add CodeSignalsSearchTool for code-level repository search

rohit-sahoo:
## Summary

Add CodeSignalsSearchTool as a fallback for searching repositories when README-based search (RepositorySearchTool) returns insufficient results. Searches actual source code content to find repos with minimal documentation.

## What have I done

- Added `CodeSignalsSearchTool` that searches LLM-extracted code signals (function names, class names, imports, code summaries)
- Implemented `_extract_summary()` to extract only "Code Summary:" sections, significantly reducing context size
- Added safety limits: max 6 results, default is 5
- Integrated with existing MCP server via `@mcp_tool` decorator

## How did I do

- Tool calls `/api/code_signals/search` endpoint with hybrid search (vector + keyword)
- Parses response and extracts summaries from `code_signals` field
- Returns `CodeSignalsHit` objects with `repo_id`, `repo_url`, `content`, and `score`

## How to test
```bash
# Run tests
uv run pytest tests/tools/code_search/test_code_signals_search.py
```

### DIFF
--- a/akd_ext/tools/__init__.py
+++ b/akd_ext/tools/__init__.py
@@ -8,7 +8,7 @@ from .sde_search import (
     SDESearchToolInputSchema,
     SDESearchToolOutputSchema,
 )
-from .code_search.code_signals_search import (
+from .code_search.code_signals import (
     CodeSignalsSearchInputSchema,
     CodeSignalsSearchOutputSchema,
     CodeSignalsSearchTool,
--- a/akd_ext/tools/__init__.py
+++ b/akd_ext/tools/__init__.py
@@ -8,6 +8,12 @@ from .sde_search import (
     SDESearchToolInputSchema,
     SDESearchToolOutputSchema,
 )
+from .code_search.code_signals_search import (
+    CodeSignalsSearchInputSchema,
+    CodeSignalsSearchOutputSchema,
+    CodeSignalsSearchTool,
+    CodeSignalsSearchToolConfig,
+)
 from .code_search.repository_search import (
     RepositorySearchTool,
     RepositorySearchToolInputSchema,
@@ -24,6 +30,10 @@ __all__ = [
     "SDESearchToolOutputSchema",
     "SDESearchToolConfig",
     "SDEDocument",
+    "CodeSignalsSearchInputSchema",
+    "CodeSignalsSearchOutputSchema",
+    "CodeSignalsSearchTool",
+    "CodeSignalsSearchToolConfig",
     "RepositorySearchTool",
     "RepositorySearchToolInputSchema",
     "RepositorySearchToolOutputSchema",
--- a/akd_ext/tools/code_search/code_signals.py
+++ b/akd_ext/tools/code_search/code_signals.py
@@ -7,6 +7,7 @@ Use as fallback when README-based search (RepositorySearchTool) is insufficient.
 
 import os
 from typing import Any, Literal
+from urllib.parse import urljoin
 
 import httpx
 from loguru import logger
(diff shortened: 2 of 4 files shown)


akd_ext/tools/code_search/code_signals.py
NISH1001, line 1: Change the name to just maybe code_signals.py search is implied so we can skip that suffix

akd_ext/tools/code_search/code_signals_search.py
```diff
@@ -0,0 +1,133 @@
...
+DEFAULT_CODE_SIGNALS_BASE_URL = "https://dyejsbdumgpqz.cloudfront.net/"
+
+
+class CodeSignalsSearchToolConfig(BaseToolConfig):
+    """Configuration for the Code Signals Search Tool."""
+
+    base_url: str = Field(
+        default=os.getenv("SDE_CODE_SIGNALS_URL", DEFAULT_CODE_SIGNALS_BASE_URL),
+        description="Base URL for the SDE Code Signals API",
+    )
```
NISH1001: What i'd want is actually base_url for sde itself...and then the code signal endpoint url as 2 fields
- `sde_base_url`
- `endpoint`

Or something. This way, we can just unify same SDE_BASE_URL env across all the tools to avoid setting same thing with multiple env variables.

akd_ext/tools/code_search/code_signals_search.py
```diff
@@ -0,0 +1,133 @@
...
+    async def _arun(self, params: CodeSignalsSearchInputSchema) -> CodeSignalsSearchOutputSchema:
+        """Execute Code Signals search."""
+        request_body = {
+            "search_term": params.query,
+            "search_type": self.config.search_type,
+            "page_size": params.limit,
+            "page": params.page,
+        }
+
+        logger.debug(f"Code Signals API request: {request_body}")
```
NISH1001: can we have this as

```python
if self.debug:
logger.debug(...)
```

akd_ext/tools/code_search/code_signals_search.py
```diff
@@ -0,0 +1,133 @@
...
+
+        url = f"{self.config.base_url.rstrip('/')}/api/code_signals/search"
```
NISH1001: once we have sde_base_url and endpoint
we can better use urlparse urljoin instead of doing manual string processing.

akd_ext/tools/__init__.py
```diff
@@ -8,6 +8,12 @@
     SDESearchToolInputSchema,
     SDESearchToolOutputSchema,
 )
+from .code_search.code_signals_search import (
```
NISH1001: from .code_search.code_signals

NISH1001 (changes requested):

akd_ext/tools/code_search/code_signals.py
rohit-sahoo: Updated!

akd_ext/tools/code_search/code_signals_search.py
rohit-sahoo: Agreed! Updated!

akd_ext/tools/code_search/code_signals_search.py
rohit-sahoo: Added!

akd_ext/tools/code_search/code_signals_search.py
rohit-sahoo: Updated!

akd_ext/tools/__init__.py
rohit-sahoo: Fixed!

NISH1001 (approved):

PR #32, opened 2026-02-13
Add build annotated types to pass the constraints for simple types

iamsims:
# `_build_annotated_type`: Why it's needed

## Problem

The `tool_converter` builds MCP function signatures by reading `field.annotation` from Pydantic's `FieldInfo`. For **non-union types** (e.g. `int`, `float`, `str`), Pydantic strips `Annotated` wrappers and stores constraints (`ge`, `le`, etc.) in `field.metadata` instead. This means `field.annotation` returns bare `int`, and FastMCP's `TypeAdapter` generates a schema **without** `minimum`/`maximum`.

**Example** — `rows: int = Field(100, ge=1, le=1000)`:

- `field.annotation` → `int`
- MCP schema → `{"default": 100, "type": "integer"}` (constraints lost)

Union types like `int | None` are unaffected — Pydantic preserves `Annotated` for complex types.

## Fix

`_build_annotated_type` reconstructs the `Annotated` wrapper by scanning `field.metadata` for constraint objects (`Ge`, `Le`, etc.) and re-wrapping:

`int` → `Annotated[int, Field(ge=1, le=1000)]`

MCP schema now correctly emits `{"default": 100, "type": "integer", "minimum": 1, "maximum": 1000}`.

## Design decision

Field descriptions are **not** forwarded through this function. The tool's `description` (class docstring, set via `__doc__` on line 98 of `converter.py`) already contains all input/output field descriptions. Only constraints (`ge`, `gt`, `le`, `lt`, `min_length`, `max_length`, `pattern`, `multiple_of`, `strict`) are recovered, since they have no other path into the MCP schema.

### DIFF
--- a/akd_ext/mcp/converter.py
+++ b/akd_ext/mcp/converter.py
@@ -2,10 +2,46 @@
 
 from collections.abc import Callable
 from inspect import Signature, Parameter
-from typing import Awaitable, Any
+from typing import Annotated, Awaitable, Any
 
+from pydantic import Field as PydanticField
+from pydantic.fields import FieldInfo
 from fastmcp import FastMCP
 from akd.tools._base import BaseTool
+
+# FieldInfo constraint attributes that map to JSON Schema keywords
+_CONSTRAINT_ATTRS = ("ge", "gt", "le", "lt", "multiple_of", "min_length", "max_length", "pattern", "strict")
+
+
+def _build_annotated_type(field: FieldInfo) -> Any:
+    """Reconstruct an Annotated type that preserves Field constraints.
+
+    Pydantic strips Annotated wrappers for non-union types, storing constraints in
+    field.metadata instead of field.annotation. This function re-wraps the annotation
+    so FastMCP's TypeAdapter emits minimum/maximum/etc. in the JSON Schema.
+
+    Note: Field descriptions are intentionally excluded here because the tool
+    description already exposes all input/output field descriptions to the LLM.
+    """
+    field_type = field.annotation
+    field_kwargs: dict[str, Any] = {}
+
+    for attr in _CONSTRAINT_ATTRS:
+        val = getattr(field, attr, None)
+        if val is not None:
+            field_kwargs[attr] = val
+
+    # Also check metadata list for constraint objects (e.g. Ge, Le) that Pydantic stores there
+    for m in field.metadata:
+        for attr in _CONSTRAINT_ATTRS:
+            val = getattr(m, attr, None)
+            if val is not None:
+                field_kwargs.setdefault(attr, val)
+
+    if not field_kwargs:
+        return field_type
+
+    return Annotated[field_type, PydanticField(**field_kwargs)]
 
 
 def tool_converter(tool: BaseTool) -> Callable[..., Awaitable[Any]]:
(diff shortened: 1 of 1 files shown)


iamsims:
The gist is: for the non union primitive types, constraints were not exposed on the MCP level. When it is a union type it is a complex type and the constraints are exposed even with the code now. 

Reason: Pydantic makes the decision to strip Annotated wrappers for non-union types and move constraints into field.metadata and FastMCP just extract what it gets from field.annotation. 

Solution: We have to reconstruct the constraints from the field.metadata and pass it to FastMCP.

---- 

Examples for your reference 

--- 

### NON-UNION + Field()
  rows: int = Field(100, ge=1, le=1000, description="Max products to return")
  BEFORE: {"default": 100, "type": "integer"}
  AFTER:  {"default": 100, "maximum": 1000, "minimum": 1, "type": "integer"}
  FIXED


### NON-UNION + Annotated
  rows: Annotated[int, Field(ge=1, le=1000)] = Field(100, description="Max products to return")
  BEFORE: {"default": 100, "type": "integer"}
  AFTER:  {"default": 100, "maximum": 1000, "minimum": 1, "type": "integer"}
  FIXED

### NON-UNION + no constraints
  rows: int = Field(100, description="Max products to return")
  BEFORE: {"default": 100, "type": "integer"}
  AFTER:  {"default": 100, "type": "integer"}
  no change

### UNION + Field()
  sol_min: int | None = Field(None, ge=0, description="Minimum sol number")
  BEFORE: {"anyOf": [{"type": "integer"}, {"type": "null"}], "default": null}
  AFTER:  {"anyOf": [{"minimum": 0, "type": "integer"}, {"type": "null"}], "default": null}
  FIXED

### UNION + Annotated
  sol_min: Annotated[int, Field(ge=0, description="Minimum sol number")] | None = None
  BEFORE: {"anyOf": [{"description": "Minimum sol number", "minimum": 0, "type": "integer"}, {"type": "null"}], "default": null}
  AFTER:  {"anyOf": [{"description": "Minimum sol number", "minimum": 0, "type": "integer"}, {"type": "null"}], "default": null}
  no change

### UNION + no constraints
  sol_min: int | None = Field(None, description="Minimum sol number")
  BEFORE: {"anyOf": [{"type": "integer"}, {"type": "null"}], "default": null}
  AFTER:  {"anyOf": [{"type": "integer"}, {"type": "null"}], "default": null}
  no change

akd_ext/mcp/converter.py
```diff
@@ -2,11 +2,47 @@
...
-from typing import Awaitable, Any
+from typing import Annotated, Awaitable, Any
 
+from pydantic import Field as PydanticField
+from pydantic.fields import FieldInfo
 from fastmcp import FastMCP
 from akd.tools._base import BaseTool
 
+# FieldInfo constraint attributes that map to JSON Schema keywords
+_CONSTRAINT_ATTRS = ("ge", "gt", "le", "lt", "multiple_of", "min_length", "max_length", "pattern", "strict")
```
NISH1001, line 13: I think ther'es a better way to get this via pydantic itself.

NISH1001: If not, we can use this for now. I did a quick check and it seems this is happenign at runtime in pydantic itself and they don't ahve any predefined set of literals for this

NISH1001 (approved):

akd_ext/mcp/converter.py
```diff
@@ -2,11 +2,47 @@
...
+
+
+def _build_annotated_type(field: FieldInfo) -> Any:
```
NISH1001, line 16: why is the return type `Any` here?

iamsims: could be any type so

NISH1001: nvermind I figured out

PR #35, opened 2026-02-20
change to single cmr care agent workflow from multi agent workflow fo…

sanzog03:
## Summary
This changes the implementation of CMR Care Agent from a multi-agent workflow (previously, CMRCareAgent(CMRSearchAgent, OutputFormatterAgent)) to a single-agent workflow. 
P.S. The previous multi-agent implementation was done because the CMR agent that came through the CARE process + OpenAI agent builder's output was free-form text. And we had to make the output such that the Backend could parse the data.

## What it does
In the multi-agent workflow, the first agent does the heavy lifting, and the second agent just formats the data. 
We tested and figured out that a single-agent workflow, when provided with the structured output schema, would yield similar quality results as the multi-agent workflow. So, these changes facilitate the exact flow.

## How it does this
- Removes the subagents with CMR data search and Output Formatter to a single one.

## Files changed
- `akd-ext/akd_ext/agents/cmr_care.py`

## Testing
```python
import asyncio

async def main():
    agent = CMRCareAgent(CMRCareConfig(debug=True))
    question = "Can you find me datasets about sea ice?"

    async for event in agent.astream(CMRCareAgentInputSchema(query=question)):
        print(event)

    asyncio.run(main())
```

Environment variables required:
```
OPENAI_API_KEY="xxxxx"
CMR_MCP_URL
```

### DIFF
--- a/akd_ext/agents/cmr_care.py
+++ b/akd_ext/agents/cmr_care.py
@@ -10,8 +10,6 @@ Public API:
 from __future__ import annotations
 
 import os
-import uuid
-from collections.abc import AsyncIterator
 from typing import Any, Literal
 
 from agents import HostedMCPTool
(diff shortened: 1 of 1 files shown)


NISH1001 (approved):
LGTM

PR #36, opened 2026-02-20
changes to manually trace enter and exit to support human interrput followed by stream cancel

sanzog03:
## Summary

Fixes a `ValueError` during `asyncio.run()` shutdown caused by a `contextvars.Context` mismatch when the `trace()` context manager exits after a human-in-the-loop (HITL) interruption in [_stream_llm_response](cci:1://file:///Users/thapa24-mbp/Devs/akd/akd-ext/akd_ext/agents/_base.py:424:4-745:20).

## What it does

Resolves the error: `ValueError: <Token ...> was created in a different Context` that occurs when the `ask_human` tool is invoked during streaming. The `stream.cancel()` call after human tool interruption causes the OpenAI Agents SDK's `trace()` context manager to exit in a different `contextvars.Context` than where it was entered, crashing during `ContextVar.reset()`.

## How it does this

Replaces the `with trace(class_name):` context manager with **manual trace lifecycle management**:

1. **Manual entry**: `current_trace.__enter__()` is called explicitly at the start of the streaming block.
2. **Early finish on HITL path**: When `ask_human` is triggered, `current_trace.finish(reset_current=True)` is called **before** `stream.cancel()` and the `yield` — while still in the original context where the trace was created.
3. **Safe cleanup in `finally`**: `current_trace.__exit__()` is wrapped in a `try/except (ValueError, RuntimeError)` to gracefully handle the case where the trace was already finished in the HITL path or the context has changed during async cleanup.

## Files changed

- **[akd_ext/agents/_base.py](cci:7://file:///Users/thapa24-mbp/Devs/akd/akd-ext/akd_ext/agents/_base.py:0:0-0:0)** — [_stream_llm_response](cci:1://file:///Users/thapa24-mbp/Devs/akd/akd-ext/akd_ext/agents/_base.py:424:4-745:20) method in [OpenAIBaseAgent](cci:2://file:///Users/thapa24-mbp/Devs/akd/akd-ext/akd_ext/agents/_base.py:151:0-857:23): replaced `with trace()` block with manual `__enter__` / `finish` / `__exit__` lifecycle management.

## Testing

- Ran `uv run python ./akd_ext/agents/cmr_care.py` — agent executes successfully without the `ValueError` on shutdown.
- The `ask_human` flow (HITL interruption) no longer triggers unhandled exceptions during `asyncio.run()` shutdown.
- Normal (non-HITL) streaming paths continue to work as expected with trace cleanup handled by the `finally` block.

### DIFF
--- a/akd_ext/agents/_base.py
+++ b/akd_ext/agents/_base.py
@@ -470,7 +470,16 @@ class OpenAIBaseAgent[InSchema: InputSchema, OutSchema: OutputSchema](BaseAgent,
         last_partial_dict = None
         current_turn_tool_calls: list[dict[str, Any]] = []  # tool calls for current assistant turn
         current_turn_has_outputs: bool = False  # whether any tool_output seen for current turn
-        with trace(class_name):
+        """
+            Note:
+            with trace(class_name): creates a bug. So, need to manually enter the trace and exit it.
+            This happens because we need to cancel the stream after human tool interruption.
+            The stream closure followed by human interruption causes the context of trace() be change
+            from its original context where it was created. The __exit__ of with trace(): breaks when this happens.
+        """
+        current_trace = trace(class_name)
+        current_trace.__enter__()
+        try:
             stream = Runner.run_streamed(
                 self._agent,
                 input=self._to_runner_input(messages),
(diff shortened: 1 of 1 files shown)


akd_ext/agents/_base.py
```diff
@@ -727,6 +738,12 @@ async def _stream_llm_response(
                 # yield {"type": StreamEventType.COMPLETED, "output": final_output}
 
             # TODO: Check how to add reflection_prompt to the Runner after the output is generated.
+        finally:
+            # try to exit the current trace.
+            try:
+                current_trace.__exit__(None, None, None)
+            except (ValueError, RuntimeError):
```
NISH1001: add, logger.warning

NISH1001 (changes requested):

akd_ext/agents/_base.py
sanzog03: resolved!

NISH1001 (approved):

PR #37, opened 2026-02-23
refactor(repo-search): compose SDE search and minify repository search output

sanzog03:
## Summary
  This PR reduces MCP/tool response payload size for `RepositorySearchTool` to control context growth and token cost.
  It refactors the tool from inheritance to composition and enforces a strict minified output schema.

  ## What it does
  - Refactors `RepositorySearchTool` to compose `SDECodeSearchTool` instead of extending it.
  - Minifies each result item to only:
    - `url`
    - `full_text`
    - `reliability_score`
    - `repository_metadata`
  - Removes legacy/heavy response fields from repository search output (`title`, `query`, `content`, `extra`, etc.).
  - Preserves GitHub metadata enrichment and reliability scoring behavior.
  - Adds stricter GitHub URL parsing for metadata enrichment.

  ## How it does this
  - `RepositorySearchTool` now extends `BaseTool` and internally instantiates `SDECodeSearchTool` with mapped config.
  - SDE result `content` is explicitly mapped to `full_text` in the minified response.
  - Output model is now a plain Pydantic `BaseModel` to avoid framework-added `_response` payload noise.
  - Repository metadata + reliability score are computed as before using existing utility functions.
  - Tests now assert exact output keys and confirm removed fields are not present.

  ## Files changed
  - `akd_ext/tools/code_search/repository_search.py`
  - `tests/tools/code_search/test_repository_search.py`

  ## Testing
  - Command:
    - `uv run pytest tests/tools/code_search/test_repository_search.py -q`
  - Result:
    - `8 passed`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -198,3 +198,6 @@ coordination/orchestration/*
 claude-flow
 # Removed Windows wrapper files per user request
 hive-mind-prompt-*.txt
+
+AGENTS.md
+SYNOPSIS.md
--- a/akd_ext/tools/code_search/repository_search.py
+++ b/akd_ext/tools/code_search/repository_search.py
@@ -1,14 +1,15 @@
-import os
 import asyncio
-from pydantic import Field, model_validator
+import os
 from urllib.parse import urlparse
-from loguru import logger
 
+from loguru import logger
+from pydantic import AnyUrl, BaseModel, Field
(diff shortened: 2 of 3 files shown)


NISH1001:
Since we removed code search tool from akd-core. i think this PR will have to be revisited...

PR #51, opened 2026-03-30
Add PDFParserTool

lpayne-dev:
Description:

## Summary
Adds a new `PDFParserTool` to `akd-ext` that parses PDF files from URLs or local paths using AKD core scraper backends, with mode-based routing (`fast`, `accurate`, `ocr`). This includes MCP registration/export wiring and focused tests for backend selection, error handling, and path normalization.

## What it does
Introduces 1 new tool under `akd_ext/tools/`:
- **PDF Parser Tool** (`akd_ext/tools/pdf_parser.py`): Parses a PDF into LLM-ready text plus metadata.

Adds typed schemas for the tool:
- `PDFParserToolInputSchema`: supports `url_or_path`, `mode`, optional `backend_hint`, and `return_format` hint (`markdown`/`html`/`json`).
- `PDFParserToolOutputSchema`: returns normalized `content` and `metadata`.

Implements backend behavior:
- `fast` defaults to `akd_simple` (`SimplePDFScraper`).
- `accurate` and `ocr` default to `akd_docling` (`DoclingScraper`) with config tuned per mode.
- Optional `backend_hint` can explicitly select backend.

Adds local-path normalization logic:
- Handles URL/file URI inputs directly.
- Normalizes local paths; preserves Windows local path behavior for scraper compatibility.

Registers and exports tool APIs:
- Adds `PDFParserTool`, `PDFParserToolInputSchema`, and `PDFParserToolOutputSchema` to `akd_ext/tools/__init__.py`.
- Tool is discoverable via MCP registry through `@mcp_tool`.

Adds tests at `tests/tools/test_pdf_parser.py` covering:
- Default routing for `fast` -> `akd_simple`
- Default routing for non-`fast` -> `akd_docling`
- Unsupported backend error path
- MCP registry inclusion
- Windows local path normalization

## How it does this
- Follows existing `BaseTool` + `InputSchema`/`OutputSchema` conventions used in `akd-ext`.
- Wraps AKD core scraper implementations (`SimplePDFScraper`, `DoclingScraper`) behind one consistent tool interface.
- Applies mode-to-backend routing with optional explicit backend override.
- Normalizes scraper output into a stable response contract: `content` + enriched `metadata` (`backend`, `return_format`, parser metadata).

## Testing
- `uv run pytest tests/tools/test_pdf_parser.py`
- `uv run pre-commit run --all-files` (ruff/format/lint checks)

### DIFF
--- a/akd_ext/tools/pdf_parser.py
+++ b/akd_ext/tools/pdf_parser.py
@@ -1,0 +1,113 @@
+"""PDF parser tool using AKD core backends."""
+
+from __future__ import annotations
+
+import sys
+from pathlib import Path
+from typing import Any, Literal
+from akd._base import InputSchema, OutputSchema
+from akd.tools import BaseTool
+from akd.tools.scrapers import (
+    DoclingScraper,
+    DoclingScraperConfig,
+    ScraperToolOutputSchema,
+    SimplePDFScraper,
+)
+from pydantic import Field
+
+from akd_ext.mcp import mcp_tool
+
+Mode = Literal["fast", "accurate", "ocr"]
+BackendHint = Literal[
+    "akd_simple",
+    "akd_docling",
+]
+
+
+class PDFParserToolInputSchema(InputSchema):
+    """Input schema for PDF parsing."""
+
+    url_or_path: str = Field(..., description="HTTP(S) URL or local filesystem path to a PDF")
+    mode: Mode = Field(default="accurate", description="Parsing mode: fast, accurate, or ocr")
+    backend_hint: BackendHint | None = Field(
+        default=None,
+        description="Optional backend override (akd_simple or akd_docling)",
+    )
+    return_format: Literal["markdown", "html", "json"] = Field(
+        default="markdown",
+        description="Preferred output format hint for backend parsing",
+    )
+
+
+class PDFParserToolOutputSchema(OutputSchema):
+    """Output schema for parsed PDF content."""
+
+    content: str = Field(..., description="Parsed text content")
+    metadata: dict[str, Any] = Field(default_factory=dict, description="Parser and document metadata")
+
+
+def _normalize_url_or_path(url_or_path: str) -> str:
+    lower = url_or_path.lower()
+    if lower.startswith(("http://", "https://", "file://")):
+        return url_or_path
+
+    p = Path(url_or_path).expanduser().resolve()
+    as_uri = p.as_uri()
+    local_path = str(p)
+
+    if sys.platform.startswith("win"):
+        return local_path
+    return as_uri
+
+async def _run_akd_simple(url_or_path: str) -> ScraperToolOutputSchema:
+    scraper = SimplePDFScraper()
+    params = scraper.input_schema(url=_normalize_url_or_path(url_or_path))
+    return await scraper._arun(params)
+
+
+async def _run_akd_docling(url_or_path: str, mode: Mode) -> ScraperToolOutputSchema:
+    if mode == "fast":
+        cfg = DoclingScraperConfig(pdf_mode="fast", do_table_structure=False, use_ocr=False)
+    elif mode == "accurate":
+        cfg = DoclingScraperConfig(pdf_mode="accurate", do_table_structure=True, use_ocr=False)
+    else:
+        cfg = DoclingScraperConfig(pdf_mode="accurate", do_table_structure=True, use_ocr=True)
+
+    scraper = DoclingScraper(cfg)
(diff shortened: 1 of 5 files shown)


NISH1001:
@lpayne-dev Could you update the body of the PR like we have for other PRs. Thanks

akd_ext/tools/pdf_parser.py
```diff
@@ -0,0 +1,113 @@
...
+BackendHint = Literal[
+    "akd_simple",
+    "akd_docling",
+]
+
+
+class PDFParserToolInputSchema(InputSchema):
+    """Input schema for PDF parsing."""
+
+    url_or_path: str = Field(..., description="HTTP(S) URL or local filesystem path to a PDF")
```
sanzog03, line 30: clarify what do you mean by local file system path

lpayne-dev: I was under the initial impression that the server runs on the user's own machine, is that incorrect? 

If incorrect, we can remove the local file path option, as it is just  a holdover from my benchmarking process.

akd_ext/tools/pdf_parser.py
```diff
@@ -0,0 +1,113 @@
...
+    p = Path(url_or_path).expanduser().resolve()
+    as_uri = p.as_uri()
+    local_path = str(p)
+
+    if sys.platform.startswith("win"):
+        return local_path
+    return as_uri
+
+async def _run_akd_simple(url_or_path: str) -> ScraperToolOutputSchema:
+    scraper = SimplePDFScraper()
```
NISH1001: provide optional config and pass that as well. Since each scrpaer is also driven by config.

scraper = SimplePDFScraper(config=config)

akd_ext/tools/pdf_parser.py
```diff
@@ -0,0 +1,113 @@
...
+    params = scraper.input_schema(url=_normalize_url_or_path(url_or_path))
+    return await scraper._arun(params)
```
NISH1001: always use `arun(...)` which is the public-facing interface not `_arun(...)`.

akd_ext/tools/pdf_parser.py
```diff
@@ -0,0 +1,113 @@
...
+        metadata = result.get("metadata", {})
+        if not isinstance(metadata, dict):
+            metadata = {"raw_metadata": metadata}
+        metadata["backend"] = backend
+        metadata["return_format"] = params.return_format
+
+        return PDFParserToolOutputSchema(
+            content=str(result.get("content", "") or ""),
+            metadata=metadata,
+        )
```
NISH1001, line 144: Should we also consider adding general scraper tool as well. akd core already has composite scraper where we can pass scraper objects. 

Maybe like `GeneralScraperTool` or something will be good addition as well and the compsoite scraper will handle url webpage, pdf, antyhing direclty as well.

NISH1001 (changes requested):

akd_ext/tools/pdf_parser.py
lpayne-dev: By local filesystem path, I mean a path that is resolvable by the AKD server process at runtime (e.g., /data/docs/a.pdf or C:\docs\a.pdf when server runs there). If the server is remote, a path on the client machine is not accessible; in that case HTTP(S) URL or accessible file:// URI should be used.

akd_ext/tools/pdf_parser.py
lpayne-dev: I’ll update both scraper paths to accept optional config and pass it through during initialization so behavior is fully config-driven.

akd_ext/tools/pdf_parser.py
sanzog03: Agreed! As the MCP server will stay away from the local path, it'll be good to reflect that.

akd_ext/tools/pdf_parser.py
```diff
@@ -0,0 +1,117 @@
...
+    input_schema = PDFParserToolInputSchema
+    output_schema = PDFParserToolOutputSchema
+
+    async def _arun(self, params: PDFParserToolInputSchema) -> PDFParserToolOutputSchema:
+        backend = params.backend_hint
+        if backend is None:
+            backend = "akd_simple" if params.mode == "fast" else "akd_docling"
+
+        if backend == "akd_simple":
+            result = _scraper_to_result(await _run_akd_simple(params.url_or_path))
```
sanzog03: config are not passed to the _run_akd_simple and _run_akd_docling while the signature allow them.
Might be a good idea to put config defination via BaseToolConfig extension and pass them if available in tool config.

PR #52, opened 2026-03-31
Add ADS Search tool

pushwithak:
## Summary

  - Adds `ADSSearchTool` — searches ADS (Astrophysics Data System) for papers and automatically extracts GitHub URLs
  from each paper's full text using the ADS highlighting API
  - Adds `ADSLinksResolverTool` — resolves external resource links (data archives, code repos, electronic sources) for a
   given ADS bibcode
  - Adds `AstroCodeSearchAgent` — orchestrates code discovery for astrophysics queries using the ADS tools alongside
  existing repository search and SDE tools
  - Both ADS tools are registered via `@mcp_tool` and exposed through the FastMCP server

  ## Key features

  - **GitHub URL extraction**: For each paper returned by ADS search, the tool runs a parallel full-text search using
  ADS highlighting to extract `github.com` URLs mentioned in the paper body, abstract, and acknowledgments
  - **Token efficiency**: Configurable abstract truncation (default 300 chars), author limiting (default 10), result
  limiting (default 5), and field presets (`minimal`, `standard`, `extended`, `full`)
  - **Field presets**: Allow the caller to control how much metadata is returned per paper, reducing context usage for
  LLM agents

### DIFF
--- a/akd_ext/agents/pds_search_care.py
+++ b/akd_ext/agents/pds_search_care.py
@@ -36,178 +36,189 @@ from loguru import logger
 
 PDS_SEARCH_AGENT_SYSTEM_PROMPT = """
 ROLE
-You are the Planetary Data Discovery Agent (NASA PDS Dataset/Product Finder). Your job is discovery and metadata only: translate a user's natural-language planetary science question into deterministic searches across NASA PDS node services/APIs, and return datasets/collections plus specific products/granules with stable identifiers + download locations + source provenance—without downloading anything.
+You are the Planetary Data Discovery Agent (NASA PDS Dataset/Product Finder).
+Your job is discovery and metadata only: translate a user's planetary-science question into bounded searches across NASA PDS discovery tools and node-operated services, then return relevant bundles/collections/datasets/products with stable identifiers and download locations when available. Do not download anything.
 
 OBJECTIVE
 Given a user query, you must:
-    Interpret scope (without inventing details).
-    Obtain required Yes/No confirmations before searching.
-    Search the appropriate PDS node services first, then broaden via PDS MCP/PDS API as a breadth pass.
-    Return results in the mandated output template with complete Candidate Dataset Metadata fields and a Search Reproducibility Log.
+1. Interpret the request without inventing facts.
+2. Ask for clarification only when the query is too ambiguous or too broad to search responsibly.
+3. Choose the right search granularity and tool type for the request.
+4. Return the strongest matching result(s) with required metadata, and include both PDS4 and PDS3 versions when available for the same underlying data or product family.
 
-CONTEXT & INPUTS
+SCOPE
+Inputs may include:
+- a natural-language planetary science query
+- optional constraints such as target, region, mission, instrument, time, resolution, geometry, processing level
+- optional prior run output for Stable vs Latest comparison
 
-    Inputs you may receive
-        User's natural-language question (may be novice → expert).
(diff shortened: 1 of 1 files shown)


akd_ext/agents/__init__.py
```diff
@@ -14,6 +14,11 @@
     CMRCareAgentOutputSchema,
     CMRCareConfig,
 )
+from akd_ext.agents.astro_code_search import (
+    AstroCodeSearchAgent,
+    AstroCodeSearchAgentInputSchema,
+    AstroCodeSearchAgentOutputSchema,
+    AstroCodeSearchAgentConfig,
```
NISH1001: We should redo the pr so that we won't be sending PR for agent. code search agent is already available in the akd_ext develop and is a generlized code search agent that works across domains. So, having astro-only is redundant and i encourage to remove it all together.

NISH1001 (changes requested):

akd_ext/tools/ads_links.py
```diff
@@ -0,0 +1,154 @@
...
+        if url and url.startswith("http"):
+            links.append(ADSLink(url=url, title="", link_type=link_type, count=count))
+
+        return links
+
+    async def _arun(self, params: ADSLinksResolverInputSchema) -> ADSLinksResolverOutputSchema:
+        """Resolve links for a given ADS bibcode."""
+        if not self.config.api_token:
+            msg = "ADS_API_TOKEN environment variable is not set. Get a token from https://ui.adsabs.harvard.edu/user/settings/token"
+            raise RuntimeError(msg)
```
NISH1001: can we maye moe this to compile-time error in init? or even in `_post_init(...)`?

```python
def _post_init():
...
   """Resolve links for a given ADS bibcode."""
super()._post_init()   
     if not self.config.api_token:
            msg = "ADS_API_TOKEN environment variable is not set. Get a token from https://ui.adsabs.harvard.edu/user/settings/token"
            raise RuntimeError(msg)


```

akd_ext/tools/ads_links.py
```diff
@@ -0,0 +1,154 @@
...
+
+        # Build URL
+        url = f"{self.config.base_url}/resolver/{params.bibcode}"
```
NISH1001: Use urllib url builder please. urljoin or something

akd_ext/tools/ads_links.py
```diff
@@ -0,0 +1,154 @@
...
+        if params.link_type:
+            url = f"{url}/{params.link_type}"
```
NISH1001: same for this. use urllib urljoin

NISH1001 (changes requested):

akd_ext/agents/__init__.py
pushwithak: Removed astro_code_search.py and its tests

akd_ext/tools/ads_links.py
pushwithak: Moved to _post_init in both ads_search.py and ads_links.py

akd_ext/tools/ads_links.py
pushwithak: Replaced with urllib.parse.urljoin

akd_ext/tools/ads_links.py
pushwithak: Replaced with urllib.parse.urljoin

PR #56, opened 2026-03-31
feat: Add descriptions to CM1 closed loop agents

sanzog03:
## Summary
This PR adds descriptive summaries to the configuration classes of all closed loop CM1 agents, detailing their roles and responsibilities within the research workflow.

## What it does
Provides better context, observability, and documentation on the exact function each agent serves by adding a detailed `description` field to their configurations. These descriptions cover the specific stages of the workflow: capability mapping, experiment design, experiment implementation, analysis/paper assembly, and report generation. 

## How it does this
By adding a default `description` string to the Pydantic configuration models (`OpenAIBaseAgentConfig` subclasses) for the following agents:
- `CapabilityFeasibilityMapperConfig`
- `WorkflowSpecBuilderConfig`
- `ExperimentImplementationConfig`
- `InterpretationPaperAssemblyConfig`
- `ResearchReportGeneratorConfig`

## Testing
- Verified that the syntax is correct and the Pydantic models initialize correctly with the new `description` default fields.
- (Note: No logical workflow changes were introduced, primarily schema documentation.)

### DIFF
--- a/akd_ext/agents/closed_loop_cm1/experiment_implementation.py
+++ b/akd_ext/agents/closed_loop_cm1/experiment_implementation.py
@@ -226,6 +226,12 @@ class ExperimentImplementationConfig(OpenAIBaseAgentConfig):
     model_name: str = Field(default="gpt-5.2")
     reasoning_effort: Literal["low", "medium", "high"] | None = Field(default="medium")
     tools: list[Any] = Field(default_factory=get_default_impl_tools)
+    description: str = Field(
+        default="Stage-4A implementation planner that translates Stage-3 workflow specs into structured "
+        "FileEdit JSON and submits experiment batches as jobs via MCP tool calls. Produces deterministic "
+        "edit definitions (namelist_param, sounding_profile, file_replace) without directly creating files "
+        "or executing commands."
+    )
 
 
 # -----------------------------------------------------------------------------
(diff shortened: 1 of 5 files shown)


akd_ext/agents/closed_loop_cm1/experiment_implementation.py
```diff
@@ -226,6 +226,12 @@ class ExperimentImplementationConfig(OpenAIBaseAgentConfig):
     model_name: str = Field(default="gpt-5.2")
     reasoning_effort: Literal["low", "medium", "high"] | None = Field(default="medium")
     tools: list[Any] = Field(default_factory=get_default_impl_tools)
+    description: str = Field(
+        default="Stage-4A implementation planner that translates Stage-3 workflow specs into structured "
+        "FileEdit JSON and submits experiment batches as jobs via MCP tool calls. Produces deterministic "
+        "edit definitions (namelist_param, sounding_profile, file_replace) without directly creating files "
+        "or executing commands."
+    )
```
muthukumaranR, line 235: use """ instead for these situations - reads better

muthukumaranR (commented):
I dont see the explanation in the definitions where the agent also can chat with the user. we'll need it to avoid FPs for the RiskAgent. unless it's addressed elsewhere

sanzog03:
> I dont see the explanation in the definitions where the agent also can chat with the user. we'll need it to avoid FPs for the RiskAgent. unless it's addressed elsewhere

@NISH1001 should we update that in the envelope schema description where it handles structuredOutput | TextOutput?
Chat with user happens after the textouput trigger??

NISH1001 (dismissed):

NISH1001 (approved):

PR #68, opened 2026-04-13
Add ADS tools for astrophysics paper and link discovery

pushwithak:
## Summary

Adds two MCP tools wrapping Astrophysics Data System (ADS) API for astrophysics paper discovery and citation-evidence retrieval.

### `ADSSearchTool`
Search the ADS literature index. Input: a query (Solr syntax supported), `rows` cap, optional `fq` filter. Output: papers with bibcode, title, authors, abstract, citation_count, DOI, publication, and names of linked data archives (HEASARC, MAST, Chandra, XMM, etc.).

### `ADSLinksResolverTool`
Given an ADS bibcode, return its "associated" bibcodes — the same records ADS shows under "Described in" on the record page. For ASCL record bibcodes (e.g. `2010ascl.soft10082F`), the associated bibcodes are the code's canonical method/description papers.

This tool exists specifically for canonical-paper recovery. ASCL's own `described_in` field is often incomplete: the FLASH ASCL record lists only the 2005 update paper (~14 citations) and misses the canonical Fryxell+ 2000 paper (~2100 citations). The ADS resolver is the only reliable way to recover the canonical paper programmatically.

## Design

- One shared `ADSToolConfig` (base_url, api_token, timeout) — both tools hit the same API.
- Per-request `httpx.AsyncClient` 
- Errors return via `output.error` instead of raising, so agents can report failures without crashing the pipeline.

## Environment

- `ADS_API_TOKEN` required. Get a token at https://ui.adsabs.harvard.edu/user/settings/token

## Testing

`tests/tools/test_ads.py` — 10 integration tests against the live ADS API (module-level skip when `ADS_API_TOKEN` is unset).

```bash
uv run pytest tests/tools/test_ads.py -n0

### DIFF
--- a/akd_ext/tools/ads.py
+++ b/akd_ext/tools/ads.py
@@ -1,0 +1,302 @@
+"""NASA Astrophysics Data System (ADS) tools.
+
+Two tools wrapping the ADS API:
+
+- ADSSearchTool:         search the literature for papers matching a query.
+- ADSLinksResolverTool:  given a bibcode, list the 'associated' bibcodes
+                          (ADS's "Described in" relationship). This is how
+                          we recover the canonical method papers for an ASCL
+                          record, which ASCL itself often under-reports.
+
+API docs: https://ui.adsabs.harvard.edu/help/api/api-docs.html
+Dev API:  https://github.com/adsabs/adsabs-dev-api
+"""
+
+from __future__ import annotations
+
+import os
+from urllib.parse import quote
+
+import httpx
+from pydantic import Field, model_validator
+
+from akd._base import InputSchema, OutputSchema
+from akd.tools import BaseTool, BaseToolConfig
+
+from akd_ext.mcp import mcp_tool
+
+
+# Fields requested from ADS. This fixed set covers the common needs for
+# scientific code discovery: identification (bibcode, doi), metadata (title,
+# authors, year, publication), relevance (abstract, citation_count), and
+# linked resources (data archive names via `data`, access via `esources`,
+# flags like REFEREED via `property`).
+_ADS_FIELDS = (
+    "bibcode,title,first_author,author,abstract,year,pubdate,"
+    "citation_count,doi,pub,data,esources,property"
+)
+
+
+# ---------------------------------------------------------------------------
+# Shared config
+# ---------------------------------------------------------------------------
+
+
+class ADSToolConfig(BaseToolConfig):
+    """Shared configuration for all ADS tools.
+
+    Both the search tool and the links-resolver tool hit the same API and
+    need the same values (base URL, bearer token, timeout), so they share
+    this single config.
+    """
+
+    base_url: str = Field(
+        default="https://api.adsabs.harvard.edu/v1",
+        description="Base URL for the ADS API.",
+    )
+
+    api_token: str = Field(
+        default_factory=lambda: os.environ.get("ADS_API_TOKEN", ""),
+        description="ADS API bearer token. Defaults to the ADS_API_TOKEN env var.",
+    )
+
+    timeout: float = Field(
+        default=30.0,
+        description="HTTP request timeout in seconds.",
+    )
+
+    @model_validator(mode="after")
+    def _require_api_token(self) -> "ADSToolConfig":
+        # ADS requires authentication. Fail fast at config creation if no token
+        # is available, rather than surfacing a 401 at query time.
+        if not self.api_token:
+            raise ValueError(
+                "ADS_API_TOKEN environment variable is not set. "
+                "Get a token from https://ui.adsabs.harvard.edu/user/settings/token"
+            )
+        return self
+
+
+# ---------------------------------------------------------------------------
+# ADSSearchTool
+# ---------------------------------------------------------------------------
+
+
+class ADSPaper(OutputSchema):
+    """A single paper returned by ADS search."""
+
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by pushwithak** [comment]
@NISH1001

**Comment by pushwithak** [comment]
@NISH1001

File: akd_ext/tools/_helpers.py [hunk 1]
```diff
@@ -0,0 +1,115 @@
...
+
+
+async def get_with_retry(
+    client: httpx.AsyncClient,
+    url: str,
+    *,
+    params: dict[str, Any] | None = None,
+    headers: dict[str, str] | None = None,
+    max_retries: int = 3,
+    base_backoff: float = 0.5,
```
NISH1001: de we actually need this? Won't it be solved via `https://github.com/jd/tenacity`?

akd_ext/tools/ads.py
```diff
@@ -0,0 +1,616 @@
...
+        default=None,
+        description=(
+            "Field preset for token efficiency. "
+            "'minimal': bibcode, title, first_author, year, citations. "
+            "'standard': + authors, date, DOI, journal, abstract, data. "
+            "'extended': + volume, page, keywords, esources, property. "
+            "'full': all fields including identifiers and affiliations. "
+            "If None, uses 'standard'."
+        ),
+    )
```
NISH1001: Should this be part of config actually? If input, LLM has more tries to change preset, right? Is the design that way currently where the LLM will choose these presets?

akd_ext/tools/ads.py
```diff
@@ -0,0 +1,616 @@
...
+    fl: str | None = Field(
+        default=None,
+        description="Custom comma-separated fields to return. Overrides field_preset if provided.",
```
NISH1001: Should this be part of config as well?

akd_ext/tools/ads.py
```diff
@@ -0,0 +1,616 @@
...
+    )
+    sort: str = Field(default="score desc", description="Sort order (e.g., 'score desc', 'citation_count desc')")
```
NISH1001: Same. Probalby part of config is better

akd_ext/tools/ads.py
```diff
@@ -0,0 +1,616 @@
...
+    fq: str | None = Field(default=None, description="Filter query to narrow results (e.g., 'property:refereed')")
+    fetch_github_urls: bool = Field(
+        default=False,
+        description=(
+            "If True, issue a secondary highlight query per result to extract GitHub URLs "
+            "from the full text. Costs 1 extra ADS quota unit per returned paper. Off by default."
+        ),
+    )
```
NISH1001: Put as part of config. We don't want LLM to figure out all these params.

akd_ext/tools/ads.py
```diff
@@ -0,0 +1,616 @@
...
+    truncate_abstract: int | None = Field(
```
NISH1001: Config

akd_ext/tools/ads.py
```diff
@@ -0,0 +1,616 @@
...
+        default=None,
+        description="Per-call override for abstract truncation (0 = no truncation). Falls back to config default.",
+    )
+    max_authors: int | None = Field(
```
NISH1001: Config

NISH1001 (changes requested):

akd_ext/tools/_helpers.py
pushwithak: done, rewrote _helpers.py using tenacity.

akd_ext/tools/ads.py
pushwithak: Updated all the above things to config.

akd_ext/tools/_helpers.py
```diff
@@ -0,0 +1,137 @@
...
+
+
+# ---------------------------------------------------------------------------
+# Tenacity retry helpers
+# ---------------------------------------------------------------------------
+
+
+def _is_retryable_status(response: httpx.Response) -> bool:
+    """Return True if the response status code is retryable."""
+    return response.status_code in _RETRY_STATUS
```
NISH1001: This can be a lambda function tiself probably.

akd_ext/tools/_helpers.py
```diff
@@ -0,0 +1,137 @@
...
+    Retries on connection errors and on status codes in ``_RETRY_STATUS``.
+    Honors ``Retry-After`` (seconds form) when present; otherwise uses
+    exponential backoff with jitter. The final response is returned even on
+    failure — the caller raises via ``raise_for_status()``.
+    """
+
+    @retry(
+        retry=retry_any(
+            retry_if_exception_type((httpx.ConnectError, httpx.ReadError, httpx.RemoteProtocolError)),
+            retry_if_result(_is_retryable_status),
```
NISH1001: maybe just lambda is okay to have here

akd_ext/tools/_helpers.py
NISH1001, line 1: the helpers might need a little bit of rework probably. Just wanted to make sure we're not introducing any claude-bloat to things that might not needed

NISH1001 (changes requested):

NISH1001 (changes requested):
@pushwithak let's simplify this and no helper. Avoid any complicated retry etc

PR #69, opened 2026-04-13
Add ASCL tools for astrophysics code discovery

pushwithak:
## Summary

Adds an MCP tool wrapping the Astrophysics Source Code Library (ASCL) for astrophysics code discovery.

### `ASCLSearchTool`
Search ASCL by name, capability keyword, or ASCL id. Returns each code's ascl_id, title, authors (credit), abstract, all associated URLs (site_list), ADS bibcode of the ASCL record itself, ADS URLs for describing papers and papers that used the code, usage count, and view count.

The tool auto-detects id-shaped queries (`1303.002`, `ascl:1303.002`) and formats them as exact-match lookups, so one tool covers both name search and id lookup.

## API quirks handled

- **PHP-serialized list fields**: `site_list`, `described_in`, `used_in` come back as PHP-serialized strings — parsed via `_parse_php_array`.
- **Query quoting**: The ASCL API requires `q` to be wrapped in quotes — returns 404 otherwise.
- **`views` type**: comes back as a string; converted to int with a fallback to 0.

## Design

- Config: `base_url` (env override via `ASCL_API_URL`), `timeout`.
- Per-request `httpx.AsyncClient` — matches the pattern of `sde_search` and the eie tools.
- Errors returned via `output.error` instead of raising.
- No shared helpers, no retry logic, no auth (ASCL is public).

## Testing

`tests/tools/test_ascl.py` — 10 integration tests against the live ASCL API.

```bash
uv run pytest tests/tools/test_ascl.py -n0

### DIFF
--- a/akd_ext/tools/__init__.py
+++ b/akd_ext/tools/__init__.py
@@ -20,6 +20,13 @@ from .code_search.repository_search import (
     RepositorySearchToolOutputSchema,
     RepositorySearchToolConfig,
 )
+from .ascl import (
+    ASCLEntry,
+    ASCLSearchTool,
+    ASCLSearchToolConfig,
+    ASCLSearchToolInputSchema,
+    ASCLSearchToolOutputSchema,
+)
 
 __all__ = [
     "DummyTool",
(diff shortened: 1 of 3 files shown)


NISH1001:
@pushwithak this also has ADS as well? Could be just ASCL only PR.

pushwithak:
It was dependent since ascl was built on ads (it was using the same _helpers.py). I updated it as ASCL only.

pushwithak:
Updated. This has only ascl now.

PR #72, opened 2026-04-17
Fix stale tests: file resolver + workflow spec builder schemas

NISH1001:
## Summary
- `test_files.py`: URLFileResolver emits `input_text`/`input_image` types (OpenAI Responses API format). Tests still expected the old `text` type. Updated to match.
- `test_files.py`: Image resolve now returns 2 parts (image + caption text). Updated assertion to `len(parts) == 2`.
- `test_workflow_spec_builder.py`: `WorkflowSpecBuilderInputSchema` requires `stage_2_feasibility` — test didn't pass it. Added a realistic default. Also dropped the obsolete `model_name` parametrize which isn't on the schema anymore.

## Test plan
- [x] `uv run pytest tests/test_files.py` — 11 passed
- [x] `uv run pytest "tests/agents/research_partner/test_workflow_spec_builder.py::test_workflow_spec_builder_agent[RQ-001: Does increasing surface roughness length affect boundary layer depth in tropical cyclones?]"` — passes (real LLM call, 24s)

### DIFF
--- a/tests/agents/research_partner/test_workflow_spec_builder.py
+++ b/tests/agents/research_partner/test_workflow_spec_builder.py
@@ -11,10 +11,20 @@ from akd_ext.agents.closed_loop_cm1 import (
 )
 
 
+_DEFAULT_FEASIBILITY = (
+    "# Feasibility Report\n\n"
+    "**Model**: CM1\n"
+    "**Capability score**: 0.85 (high)\n\n"
+    "The proposed experiment is feasible with the chosen model. "
+    "Required parameters are available in the model configuration."
+)
+
+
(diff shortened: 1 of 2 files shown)


tests/agents/research_partner/test_workflow_spec_builder.py
rohit-sahoo, line 1: Could we remove this test case, its wasting tokens running the actual LLM call

rohit-sahoo (changes requested):

rohit-sahoo (approved):
Approved! 🚢 .. we will remove the LLM call based test later

PR #79, opened 2026-04-23
feat(agents): add ReflectionCapability + scalar wiring

sanzog03:
## Summary
  Adds the next scalar-to-capability mapping on `PydanticAIBaseAgent`: `reflection_prompt → ReflectionCapability`. When the config carries a reflection prompt, the agent will inject it as a system-level nudge before every model request after the first — a hook that forces the model to step back and reason about tool results before returning a final answer, mirroring the
  legacy AKD agent's reflection pass.

  ## What it does
  - Agents built on `PydanticAIBaseAgent` now respect `config.reflection_prompt` — set it to a string and the agent automatically injects that string as a `SystemPromptPart` on every model  request after the first.
  - The initial user-facing turn is preserved untouched; reflection kicks in only once there's something to reflect *on* (turn 2+).
  - Subclasses can still extend `_build_capabilities_from_scalars` via `super()` + append — the extension pattern is pinned by a test.
  - No new config fields on `PydanticAIBaseAgentConfig`; `reflection_prompt` comes in from `BaseAgentConfig` unchanged.

  ## Files changed
  - **New files:**
    - `akd_ext/agents/_base/pydantic_ai/_capabilities.py` — holds `ReflectionCapability` (Hooks factory with closured turn counter). Module docstring notes future factories land here.
  - **Modified files:**
    - `akd_ext/agents/_base/pydantic_ai/_base.py` — imports `ReflectionCapability`; `_build_capabilities_from_scalars` body replaced with the `reflection_prompt`.
    - `tests/agents/test_base_pydantic.py` — adds four tests under a new "Phase 3b: ReflectionCapability" section; adds `enable_extra_capability: bool` on `_EchoConfig` for the
  subclass-extension test.

  ## Testing
  Hermetic test suite (no network, no model keys):

      uv run pytest tests/agents/test_base_pydantic.py -n=3

### DIFF
--- a/akd_ext/agents/_base/pydantic_ai/_base.py
+++ b/akd_ext/agents/_base/pydantic_ai/_base.py
@@ -50,6 +50,7 @@ from akd._base.protocols import AKDExecutable, RunContextProtocol
 from akd._base.structures import RunContext as AKDRunContext
 from akd.agents._base import BaseAgentConfig
 
+from ._capabilities import ReflectionCapability
 from ._event_translator import pai_event_to_akd_event
 from ._tool_adapter import akd_to_pai_tool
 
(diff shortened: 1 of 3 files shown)


NISH1001:
Closing this for future reference. Let's look at this later when needed.

PR #98, opened 2026-04-30
fix: ToolResultEvent for non MCP call giving `unknown` tool name

paridhi-parajuli:
## Summary

Fixes an issue where `ToolResultEvent.data.result.tool_name` was always `"unknown"` for non-MCP tool calls in `OpenAIBaseAgent`.

## Cause

In the `tool_output` branch, the tool name was retrieved using:

```python
getattr(raw_item, "name", "unknown")
```                                                                                                                                                                                            
But on a `tool_output` item, `raw_item`  payload doesn't carry the function name. Only the original `tool_called` raw_item does. The MCP path is unaffected.                                                                                                                                                                                             
                  
## Fix                                                                                                                                                                                                          
Look the name up by `tool_call_id` from `current_turn_tool_calls`, which the `tool_called` branch already populates. Keeps `"unknown"` as a final fallback.

### DIFF
--- a/akd_ext/agents/_base/openai.py
+++ b/akd_ext/agents/_base/openai.py
@@ -776,7 +776,14 @@ class OpenAIBaseAgent[InSchema: InputSchema, OutSchema: OutputSchema](
                                 data=ToolResultEventData(
                                     result=ToolResult(
                                         tool_call_id=tool_call_id,
-                                        tool_name=getattr(raw_item, "name", "unknown"),
+                                        tool_name=next(
+                                            (
+                                                tc["function"]["name"]
+                                                for tc in current_turn_tool_calls
+                                                if tc["id"] == tool_call_id
+                                            ),
+                                            "unknown",
+                                        ),
                                         content=tool_output_content,
                                     )
                                 ),


akd_ext/agents/_base/openai.py
```diff
@@ -776,7 +776,14 @@ async def _run_engine_stream(
...
                                     result=ToolResult(
                                         tool_call_id=tool_call_id,
-                                        tool_name=getattr(raw_item, "name", "unknown"),
+                                        tool_name=next(
+                                            (
+                                                tc["function"]["name"]
+                                                for tc in current_turn_tool_calls
+                                                if tc["id"] == tool_call_id
+                                            ),
+                                            "unknown",
```
NISH1001, line 785: Is it an iterator? What happens if there's no match? does `next` still works

paridhi-parajuli: `current_turn_tool_calls` is a list[dict] that is accumulated as the tool_called events happen. but `(tc["function"]["name"] for tc in current_turn_tool_calls if tc["id"] == tool_call_id)` is a generator expression and that's the iterator.
If there are no matches for that `tool_call_id`, the generator yields nothing. Since I'm using next(iterator, default) the two-arg form, it returns "unknown" instead of raising `StopIteration`.

NISH1001 (approved):

PR #106, opened 2026-06-09
feat: Prithvi-EO-2.0 closed-loop 6-stage agent pipeline

rohit-sahoo:
## Summary

Implements the complete Prithvi-EO-2.0 closed-loop scientific workflow as a 6-stage conversational agent pipeline, from research gap detection through experiment execution to publication-quality report generation.

## What we did

**6 conversational agents** (all chat-based, all with system prompts + structured I/O schemas):

| Stage | Agent | Tools | What it does |
|:-----:|-------|-------|--------------|
| 1 | Gap Agent | none | Reads a research corpus, surfaces knowledge gaps, proposes ranked RQs with H0/H1 |
| 2 | Capability-Feasibility Mapper | none | Maps each RQ against Prithvi capabilities, produces Go/Conditional-Go verdicts |
| 3 | Workflow Spec Builder | geocode, reverse_geocode (MCP) | Generates experiment spec + pipeline config YAML |
| 4 | Pipeline Submission | screen_events, check_screening, job_submit (MCP) | Async event screening, user approval, config submission to HPC |
| 5 | Experiment Analysis | job_status, job_plot (MCP) | Checks job completion, fetches figures + reports, runs ImageAnalyzerAgent (vision) |
| 6 | Research Report Generator | none | Produces publication-quality scientific report from all upstream outputs |

**Files changed:**
- akd_ext/agents/closed_loop/prithvi/ — agents, prompts, tools, context files, orchestrator
- akd_ext/agents/closed_loop/stages/ — experiment_analysis, experiment_implementation, research_report_generator, workflow_spec_builder
- akd_ext/agents/image_analyzer.py — added inference field, removed context dump from markdown output
- akd_ext/tools/ — geocode, reverse_geocode (new), __init__.py
- tests/agents/closed_loop/prithvi/test_prithvi_stages.py — 29 tests covering all stages

## How we did it

- **Stages 1-2**: Pure LLM reasoning, no tools. SME-designed prompts with minor allowed variations (file renames, H0/H1 preservation, 20-event hard limit, config-not-in-display, stage-complete signal, metrics inclusion).
- **Stage 3**: Generates workflow spec + config YAML. Has `geocode` and `reverse_geocode` MCP tools available for resolving user-provided place names or bounding boxes to states/coordinates when specifying custom events.
- **Stage 4**: Async screening via MCP (screen_events → check_screening poll → present events → user approves → job_submit). Config strict structure enforced by prompt — only events section changes, rest untouched.
- **Stage 5**: Deterministic orchestrator (_arun override) for traceability — calls job_status/job_plot via MCP, delegates vision to ImageAnalyzerAgent. Status-gated: analysis never fires until job is complete.
- **Stage 6**: No tools. Receives figure analysis + workflow spec, produces report with inline figures. No future-directions section (SME decision).

## How to test

**Prerequisites:**
- Set env vars: PRITHVI_MCP_URL, PRITHVI_MCP_API_KEY, OPENAI_API_KEY

**Run tests:**
```bash
python -m pytest tests/agents/closed_loop/prithvi/ -v

### CODE REVIEW

File: akd_ext/agents/closed_loop/prithvi/orchestrator.py
Comment (sanzog03) [inline akd_ext/agents/closed_loop/prithvi/orchestrator.py:1]: delete this orchestrator

File: akd_ext/agents/closed_loop/stages/research_report_generator.py [hunk 1]
```diff
@@ -49,23 +49,36 @@ class ResearchReportGeneratorConfig(ClosedLoopStageConfig):
...
         description=(
             "Stage 3 workflow specification markdown containing: research question, "
             "hypothesis, control definition, experiment matrix, feasibility notes, "
-            "and feasibility summary. This is the primary source of scientific context."
+            "and feasibility summary."
         ),
     )
-    job_id: str = Field(
-        ...,
+    figure_analysis: str = Field(
```
sanzog03: "summary": "Required job_id removed from the shared ResearchReportGeneratorInputSchema, breaking the unchanged CM1 report agent whose prompt mandates 'You receive a
  single job_id from the input ... Call job_status(job_id=...)' — callers' job_id is now silently dropped (extra=ignore)",
      
"failure_scenario": "CM1 pipeline runs Stage 6: existing callers pass job_id, pydantic silently drops it, the CM1 agent has no job_id to satisfy its mandatory Step 1
  → it stalls ('STOP immediately') or hallucinates a job_id and fetches wrong/no figures. Verified: cm1/ is untouched by this PR and the field is gone."

akd_ext/agents/closed_loop/stages/experiment_analysis.py
```diff
@@ -150,18 +292,32 @@ async def _mcp_call(
...
+
+    Handles two common server wrappers:
+    - FastMCP cloud wraps responses as ``{"result": "<json_string>"}`` — the
+      inner JSON string is parsed and returned directly.
+    - Servers that wrap the real payload as a JSON string under ``"payload"``
+      — that gets unwrapped into ``"payload_parsed"``.
     """
     timeout = httpx.Timeout(connect=30.0, read=read_timeout, write=60.0, pool=60.0)
-    async with httpx.AsyncClient(headers={"X-API-Key": api_key}, timeout=timeout) as http_client:
+    async with httpx.AsyncClient(headers={"Authorization": f"Bearer {api_key}"}, timeout=timeout) as http_client:
```
sanzog03: "summary": "Shared _mcp_call auth header changed from 'X-API-Key' to 'Authorization: Bearer' for Prithvi, but the CM1 ExperimentAnalysisAgent inherits the same
  helper while cm1/tools.py still sends X-API-Key to the same CM1 server",

"failure_scenario": "PLAUSIBLE (depends on server config): if the CM1 MCP server expects X-API-Key as cm1/tools.py suggests, every CM1 Stage-5 job_status/job_plot
  call now gets 401 → 'Could not reach MCP server' for all CM1 jobs. The auth scheme is a per-server property and should live in config/an overridable hook, not be flipped
  in the shared function."

akd_ext/agents/closed_loop/stages/experiment_analysis.py
```diff
@@ -118,19 +239,40 @@ class ExperimentAnalysisInputSchema(InputSchema):
...
+    generate a narrative report — that is Stage 6's job.
+
+    Downstream consumers (Stage 6 / Paper Assembly) receive:
+    - ``analyses``: per-figure ``FigureAnalysis`` objects from ImageAnalyzerAgent
+    - ``text_content``: concatenated CSV tables / markdown from pipeline outputs
+    - ``experiment_map``: ``{experiment_id: [figure_slugs]}`` for cross-referencing
     """
 
-    __response_field__ = "markdown"
+    __response_field__ = "analyses"
```
sanzog03: "summary": "__response_field__ changed from 'markdown' (str) to 'analyses' (list[dict]), violating the upstream _response computed field's '-> str' contract on the
  shared output schema",

"failure_scenario": "Verified live: model_dump_json() emits PydanticSerializationUnexpectedValue and serializes _response as a JSON array; any consumer rendering the
  response channel gets a raw list of dicts instead of the markdown report (which still exists in .markdown, the previous response field)."

akd_ext/agents/closed_loop/stages/experiment_analysis.py
```diff
@@ -62,6 +62,127 @@
...
+            image_urls.append(url)
+        elif kind in ("csv", "text"):
+            try:
+                text = await _fetch_text(url)
+                text_parts.append(text)
+                logger.info(f"[Stage5] fetched {kind} content from {url} ({len(text)} chars)")
+            except Exception as exc:
+                logger.warning(f"[Stage5] failed to fetch {url}: {exc}")
+                text_parts.append(f"*Could not fetch `{url}`: {exc}*")
+        else:
```
sanzog03, line 196: "summary": "Unknown-extension URLs are fetched as text with an image fallback only on exception — but httpx resp.text never raises on binary content, so the fallback
  is unreachable on HTTP 200",

"failure_scenario": "job_plot returns an extensionless image URL (or .pdf/.json/.nc artifact): the GET succeeds, PNG bytes are decoded as mojibake and appended to
  text_content (fed to Stage 6's LLM context), and the figure never reaches ImageAnalyzerAgent. Pre-PR code sent every URL to the image analyzer. Content-Type from the
  response headers is the right classifier."

akd_ext/agents/closed_loop/prithvi/tools.py
```diff
@@ -0,0 +1,128 @@
...
+    Note: the local ``screen_events``/``geocode``/``reverse_geocode`` function
+    tools still exist in this module but are not attached to the agent — event
+    screening runs on the MCP server so the slow NOAA/HLS crawl doesn't block
+    the agent loop.
+    """
+    tools: list[OpenAITool] = []
+
+    api_key = os.environ.get("PRITHVI_MCP_API_KEY")
+    if api_key:
+        url = os.environ.get("PRITHVI_MCP_URL", "")
```
sanzog03: "summary": "All three get_default_*_tools guard only on PRITHVI_MCP_API_KEY; PRITHVI_MCP_URL defaults to '' and is passed straight through as server_url (also
  agents.py:201 and cm1/tools.py:55 EXPERIMENT_STATUS)",

"failure_scenario": "Key set but URL unset → HostedMCPTool built with server_url='' → every agent run fails at the LLM API with an invalid-server-URL error instead
  of degrading to no tools like the missing-key path. Unlike cm1/tools.py:26 which has a real hardcoded fallback URL."

akd_ext/tools/reverse_geocode.py
```diff
@@ -0,0 +1,155 @@
...
+            "Convert a bounding box [west, south, east, north] to a "
+            "location: US state code, state name, county, and display name. "
+            "Uses the bbox centroid for the lookup."
+        ),
+    )
+
+
+def _reverse_geocode(bbox: list[float]) -> dict:
+    """Call Nominatim reverse API on the bbox centroid."""
+    west, south, east, north = bbox
```
sanzog03, line 105: "summary": "'west, south, east, north = bbox' has no length validation (schema is bare list[float]) and sits outside the try block, so wrong-length input raises an
  unhandled ValueError",

"failure_scenario": "Verified live: ReverseGeoCodeInput(bbox=[1.0, 2.0]) passes validation, then _reverse_geocode raises 'not enough values to unpack' which
  propagates out of _arun (only requests.RequestException is caught) — an opaque tool crash instead of the graceful {'message': ...} error path. An LLM passing a [lon,
  lat] point is a realistic call."

akd_ext/agents/closed_loop/stages/experiment_analysis.py
```diff
@@ -62,6 +62,127 @@
...
+        for row in data:
+            # Pad/trim to header length
+            padded = row + [""] * (len(header) - len(row))
+            lines.append("| " + " | ".join(padded[: len(header)]) + " |")
+        if len(rows) - 1 > 200:
+            lines.append(f"\n*({len(rows) - 1} total rows — showing first 200)*")
+        return "\n".join(lines)
+
+    # Markdown / plain text — return as-is with a header
+    return f"### Report: `{slug}`\n\n{content}"
```
sanzog03, line 173: "summary": "_fetch_text caps CSVs at 200 rows but returns .md/.txt/.log and unknown-extension bodies completely uncapped into text_content, which feeds Stage-6 LLM
  prompts",

"failure_scenario": "A multi-MB pipeline .log (explicitly in _TEXT_EXTS) is downloaded whole and concatenated into text_content → Stage 6 prompt balloons to millions
  of tokens → context overflow or runaway cost. Cap text payloads (first N KB + truncation note) like the CSV branch already does."

akd_ext/tools/geocode.py
```diff
@@ -0,0 +1,123 @@
...
+    try:
+        resp = requests.get(
+            NOMINATIM_URL, params=params, headers=HEADERS, timeout=10
+        )
+        resp.raise_for_status()
+        results = resp.json()
+    except requests.RequestException as e:
+        return {"message": f"Geocoding service unavailable: {e}"}
+    finally:
+        time.sleep(1)  # Nominatim enforces 1 req/sec
```
sanzog03: "summary": "time.sleep(1) in finally inside asyncio.to_thread neither enforces Nominatim's 1 req/sec under concurrency nor avoids penalizing every call (duplicated
  at reverse_geocode.py:118)",

"failure_scenario": "Concurrent geocode/reverse_geocode calls (tools are stateless and run in parallel) each fire immediately and sleep afterwards in separate
  threads → Nominatim still sees a burst (risking a User-Agent/IP ban), while every call including failures pays a flat 1s after the response is already in hand. A
  module-level throttle (lock + last-request timestamp) shared by both tools is the correct mechanism."

akd_ext/agents/closed_loop/stages/experiment_analysis.py
```diff
@@ -397,23 +616,54 @@ async def _astream(
...
+            source=cn,
+            message=f"{len(image_urls)} images, {len(fetched_text)} chars text. Analyzing…",
+            run_context=run_context,
+        )
 
-        # 4. Delegate — forward ImageAnalyzerAgent's streaming events
         context = self._build_context(params, experiment_urls)
+
+        # Build experiment → slug mapping for downstream consumers
+        exp_map: dict[str, list[str]] = {}
```
sanzog03: "summary": "Cleanup: _astream's tail (exp_map building, no-image early return, output construction) duplicates _arun lines 537-580 verbatim, and the slug expression
  url.rstrip('/').split('/')[-1].split('?')[0] appears 4 times (lines 107, 402, 555, 636)",

"failure_scenario": "Verified verbatim duplication: the next output field or slug tweak lands in one path and not the other, so streaming and non-streaming runs
  return different payloads to Stage 6. image_analyzer.py already established the shared-helper pattern (_process_batches used by both paths); extract _slug() and a shared
  output-builder."

sanzog03 (commented):
Minor things I dropped for the cap: the triplicated HostedMCPTool construction in prithvi/tools.py (a small _prithvi_mcp_tool(allowed_tools, description) factory
  collapses it), reverse_geocode.py duplicating geocode.py's entire Nominatim plumbing, the 51-entry hardcoded US-state table when Nominatim already returns
  ISO3166-2-lvl4, per-URL httpx.AsyncClient creation with sequential fetching in _categorize_and_fetch, and prithvi context .md files being re-read from disk on every
  config instantiation (a cached _read(name) helper fixes it).

rpelevin:
Leaving the small fixture/checklist version here since this PR is where Stage 4 wires `screen_events -> check_screening -> user approval -> job_submit`.

I would make the conformance boundary the final config-approval -> job_submit handoff:

- compute a stable digest/ref for the effective Stage 3 config after event-selection edits, before asking "All good to submit?";
- record the explicit final config approval against that digest/ref, selected event refs, run/workflow context, user/approver, decided_at, and optional expiry;
- immediately before `job_submit`, recompute the digest for the payload being submitted and require it to match the approved digest/ref;
- if events, output target, Prithvi task flags, baselines, analysis tests, or any other effective config field changed after approval, require fresh approval;
- pending, rejected, expired, or failed-validation states must not be able to call `job_submit`;
- resume after approval should submit the same payload digest and record returned `job_id` / `workspace_name` plus the pipeline's `config_received` as outcome evidence;
- duplicate flood/burn events from the same real-world event should stay visible in the approval payload before submission.

Minimal tests I would add around Stage 4:

1. approval of event selection alone cannot authorize `job_submit`;
2. final config approval over digest `A` allows exactly the payload with digest `A`;
3. mutating `events.custom_events`, `output.dir`, task flags, baselines, or analysis tests after approval blocks `job_submit`;
4. pending, rejected, expired approval and failed validation all fail closed before the MCP call;
5. `config_received` / `job_id` / `workspace_name` link back to the pre-submit approval record.

The point is not another approval layer; it is making the existing three-step enforcement reproducible at the consequential boundary before HPC execution.

akd_ext/agents/closed_loop/stages/research_report_generator.py
rohit-sahoo: Fixed. Restored it to workflow_spec + job_id and gave Prithvi its own input schema with input_schema override, since Prithvi Stage 6 gets pre-analyzed figures from Stage 5 instead of a job_id.

akd_ext/agents/closed_loop/stages/experiment_analysis.py
rohit-sahoo: Fixed. Moved the auth scheme to an overridable _auth_headers property: generic/CM1 keeps X-API-Key, Prithvi overrides to Bearer. _mcp_call now just takes a headers dict not in the shared helper.

akd_ext/agents/closed_loop/stages/experiment_analysis.py
rohit-sahoo: Fixed. Pointed it back to markdown (the rendered figure-analysis string); analyses stays as a structured field that Stage 6 reads directly. Verified model_dump_json() no longer emits the serialization warning and _response is a string.`

akd_ext/agents/closed_loop/stages/experiment_analysis.py
rohit-sahoo: Fixed

akd_ext/agents/closed_loop/prithvi/tools.py
rohit-sahoo: Fixed. The factory now guards on both PRITHVI_MCP_API_KEY and PRITHVI_MCP_URL; if either is missing it returns no tools instead of building a HostedMCPTool with server_url=''. (The earlier refactor already collapsed all three get_default_*_tools into one _prithvi_mcp_tool factory, so this is a one-place fix.)

akd_ext/tools/reverse_geocode.py
rohit-sahoo: Fixed. Added a length check at the top of _reverse_geocode: a bbox that isn't 4 elements now returns the graceful {"message": ...} error instead of an unhandled ValueError on the unpack. Verified bbox=[1.0, 2.0] returns a clean message. (Same guard will go on the MCP server's copy.)

akd_ext/agents/closed_loop/stages/experiment_analysis.py
rohit-sahoo: Fixed

akd_ext/tools/geocode.py
rohit-sahoo: Fixed

akd_ext/agents/closed_loop/stages/experiment_analysis.py
rohit-sahoo: Fixed

rpelevin:
Rohit, the review-fix commit looks like it tightened the surrounding Prithvi stage code well.

One remaining conformance boundary I would keep separate from the prompt instruction is the `job_submit` authority record. The Stage 4 prompt now says to present screened events, stop, ask "All good to submit?", and only then call `job_submit`. That is the right operator flow. The acceptance test I would still add is for the boundary underneath that flow:

1. `screen_events` returns a `task_id`;
2. `check_screening(task_id)` returns the candidate event set;
3. the approved event set is canonicalized into the final config and bound to a digest;
4. `job_submit` is allowed only for that approved config digest;
5. changing event ids, bbox/date fields, crop dates, `max_events`, hazard type, output workspace, or pipeline parameters after approval fails closed before `job_submit`;
6. the submission output carries `job_id`, `workspace_name`, the approved config digest, and the approval decision id / timestamp.

The reason I would test this separately is that the current HostedMCPTool config still exposes `screen_events`, `check_screening`, and `job_submit` together with `require_approval: "never"`. That can be fine for this PR if the prompt is the first slice, but the conformance test should make clear that approval is not just "the model remembered to ask"; it is an approved screening snapshot plus exact config digest that the `job_submit` call must match.

akd_ext/agents/closed_loop/stages/experiment_analysis.py
```diff
@@ -317,16 +548,66 @@ async def _fetch_plot_urls(self, job_id: str, workspace: str, user: str) -> dict
...
+            exp_id: [_slug(u) for u in urls if isinstance(u, str)]
+            for exp_id, urls in experiment_urls.items()
+        }
+
+    @staticmethod
+    def _text_only_output(
+        fetched_text: str, exp_map: dict[str, list[str]]
+    ) -> ExperimentAnalysisOutputSchema:
+        """Output for the no-image case (text artifacts only, no figures)."""
+        return ExperimentAnalysisOutputSchema(
```
sanzog03, line 595: "summary": "_text_only_output never sets markdown, but __response_field__ now points at 'markdown' — the no-image path returns a blank response",


"failure_scenario": "CONFIRMED (executed): job_plot returns only CSV/markdown artifacts, no figures. The output has text_content populated but _response == ''. Any
  chat/UI surface rendering the response channel shows an empty reply for a successful stage. 

Fix: set markdown=fetched_text (or the fallback message) in
  _text_only_output."

rohit-sahoo: fixed. _text_only_output now sets markdown (= fetched text, or the fallback message), so a figure-less run no longer returns a blank response.

sanzog03 (approved):

PR #108, opened 2026-06-11
feat: Add code generation pipeline with designer, generator, and Validator

rohit-sahoo:
## Summary

Adds a generic code generation pipeline that turns a research hypothesis and experiment design into a validated, self contained Python analysis script. Domain knowledge is injected through config only, no pipeline code changes are needed to support a new domain. 

CM1 is validated end-to-end on two research questions with SME-reviewed figures.

## What did we do

- **CodeDesignerAgent**: produces a markdown design document from the hypothesis + experiment context: goal & success criteria, workflow, constraints, I/O contract, and visualisation specifications (every figure must help answer the hypothesis; causal-chain coverage sets the floor)
- **CodeGeneratorAgent**: writes a complete single-file Python script from the design, with a strict two-argument CLI (`--input-dir` / `--output-dir`) and all parameters as named constants
- **Code Validator (3 layers)** —
  1. AST allowlist: 30 approved imports, blocks the four dynamic-execution
     builtins (`eval`/`exec`/`compile`/`__import__`)
  2. Bandit static security scan, fail-closed (missing bandit or HIGH severity finding = rejection)
  3. `IntentCheckerAgent` (LLM-as-judge): verifies the code against the design's success criteria using hard (scientific) vs soft
- **CodeGeneratorPipeline**: deterministic orchestrator (never calls an LLM itself): designer → generator → validators, up to 4 retries with structured correction feedback on each failure
- **CM1 domain context**: `CM1_DATA_FORMAT_CONTEXT` and  `CM1_ANALYSIS_METHODOLOGY` 
- Added `bandit` as a dependency

## How did we do

- All sub-agents are built on `PydanticAIBaseAgent` (gpt-5.2); their `check_output` validators are wired into pydantic_ai's retry loop, so structural validation retries happen inside each agent
- The orchestration loop is implemented once in `_astream`; `_arun` drains the stream, batch and streaming consumers get identical behavior
- Validation runs cheapest-first: AST parse + allowlist (instant), then bandit, then the LLM judge; failures return structured reasons that become`correction_feedback` for the next generation attempt
- The intent checker receives the same data-format reference as the generator, so it judges reader code against documented ground truth instead of generic assumptions.
- Validated with four approved end-to-end runs across two research questions.

## How to Test

```uv run pytest tests/agents/closed_loop/ -o "addopts="```

### CODE REVIEW

File: akd_ext/agents/code_generator/generator.py [hunk 1]
```diff
@@ -0,0 +1,280 @@
...
+            "Every script must accept --input-dir and --output-dir."
+        )
+    if "--output-dir" not in code:
+        return (
+            "code is missing mandatory --output-dir argument. "
+            "Every script must accept --input-dir and --output-dir."
+        )
+
+    # Enforce exactly 2 CLI arguments — no extras allowed
+    add_arg_count = code.count("add_argument(")
```
sanzog03: "summary": "Generator output validation uses raw substring heuristics (count of 'add_argument(',
  '--input-dir' in code) that both false-reject and false-pass", 

"failure_scenario": "A docstring containing 'add_argument(' inflates the count past 2 → valid code
  rejected, burning an LLM retry; '--input-dir' appearing only in a usage comment passes code whose argparse never defines it → broken script approved"

akd_ext/agents/code_generator/code_gen.py
```diff
@@ -0,0 +1,421 @@
...
+            run_context=run_context,
+        )
+
+        # 1. Design. Structural validation + retry happen inside arun():
+        #    the agent's check_output is registered as a pydantic_ai output
+        #    validator, so invalid designs are retried before this returns.
+        yield RunningEvent(source=cn, message="Producing design document…", run_context=run_context)
+
+        designer = self._make_designer()
+        plan_output = await designer.arun(params.to_designer_input())
```
sanzog03: "summary": "Sub-agent arun() calls are unguarded; retry-exhaustion or API exceptions escape _astream
  with no CompletedEvent and bypass the correction_feedback loop", 

"failure_scenario": "Generator persistently fails its check_output → pydantic_ai raises after internal
  retries → exception propagates out of the pipeline instead of becoming correction_feedback for the next attempt; stream dies without a final event"

akd_ext/agents/code_generator/validator.py
```diff
@@ -0,0 +1,577 @@
...
+                top = alias.name.split(".")[0]
+                if top not in ALLOWED_IMPORTS:
+                    violations.append(
+                        f"line {node.lineno}: `import {alias.name}` — "
+                        f"`{top}` is not in the allowed-import list"
+                    )
+
+        # Rule 1: from x import y
+        elif isinstance(node, ast.ImportFrom):
+            top = (node.module or "").split(".")[0]
```
sanzog03: "summary": "check_allowlist silently accepts relative imports because node.module=None yields
  top='' and `if top` skips the check", 

"failure_scenario": "Generated code containing 'from . import anything' passes the allowlist unflagged (fails only later at runtime
  with ImportError); validator is fail-open for this node type"

akd_ext/agents/code_generator/code_gen.py
```diff
@@ -0,0 +1,421 @@
...
+        )
+        yield CompletedEvent(
+            source=cn, message=f"Completed {cn}",
+            data=CompletedEventData(output=final_output), run_context=run_context,
+        )
+
+    # -- Output validation -------------------------------------------------
+
+    def check_output(self, output) -> str | None:
+        if isinstance(output, CodeGeneratorPipelineOutputSchema):
```
sanzog03: "summary": "CodeGeneratorPipeline.check_output is dead code — never invoked because the pipeline
  overrides _arun/_astream and bypasses the base's output-tool machinery that calls it", 

"failure_scenario": "An approved result with empty vetted_code.code is returned to
  callers; the guard written to catch it never runs"

akd_ext/agents/code_generator/designer.py
```diff
@@ -0,0 +1,422 @@
...
+        return "design_document is empty — produce a comprehensive design spec."
+    if len(doc) < 200:
+        return (
+            "design_document is too short. Include all required sections: "
+            "GOAL & SUCCESS CRITERIA, APPROACH, WORKFLOW, CONSTRAINTS, "
+            "INPUT/OUTPUT, LIBRARIES, BEST PRACTICES, and EDGE CASES."
+        )
+
+    doc_upper = doc.upper()
+    missing = [s for s in _REQUIRED_SECTIONS if s not in doc_upper]
```
sanzog03: "summary": "Required-sections validation checks generic words as substrings of the whole upper-cased
  doc — vacuously passes and can false-fail", 

"failure_scenario": "A design doc missing an actual LIBRARIES section passes because 'libraries' appears in prose; a doc headed 'OBJECTIVE' instead of 'GOAL' is rejected despite being structurally complete"

akd_ext/agents/code_generator/code_gen.py
```diff
@@ -0,0 +1,421 @@
...
+                    failure_reason=last_reason[:200],
+                ))
+                logger.warning(f"[CodeGenPipeline] attempt {attempt} generator error")
+                yield RunningEvent(source=cn, message=f"Attempt {attempt}: generator error", run_context=run_context)
+                continue
+
+            # 3a. Deterministic validators
+            yield RunningEvent(source=cn, message=f"Running validators (attempt {attempt})…", run_context=run_context)
+
+            det = self._run_deterministic_validators(gen_output.code)
```
sanzog03: "summary": "Blocking subprocess.run (bandit, timeout=30) executes inside async _astream, freezing
  the event loop on every attempt",

"failure_scenario": "In the FastMCP-served async environment, each bandit scan (1–30s) blocks all concurrent coroutines/clients; up to
  4 retries compounds it — fix with await asyncio.to_thread(self._run_deterministic_validators, ...)"

akd_ext/agents/code_generator/designer.py
```diff
@@ -0,0 +1,422 @@
...
+    criteria, approach, workflow, constraints, I/O spec, libraries, best
+    practices, plot specs (if applicable), and edge cases. The code
+    generator follows it to produce correct code on the first attempt.
+    """
+
+    input_schema = CodeDesignerInputSchema
+    output_schema = CodeDesignerOutputSchema
+    config_schema = CodeDesignerConfig
+
+    def __init__(self, config: CodeDesignerConfig | None = None) -> None:
```
sanzog03, line 468: "summary": "Prompt-assembly pattern (_assemble_*_prompt + __init__ model_copy of system_prompt) is triplicated across designer/generator/validator, with _DATA_FORMAT_CONTEXT_DESC duplicated in two files with divergent wording",

"failure_scenario": "Any change to assembly (new section, ordering, caching) must be made and tested in three places; the field docs have already drifted between designer.py and generator.py"

akd_ext/agents/code_generator/code_gen.py
```diff
@@ -0,0 +1,421 @@
...
+            if not output.rejected and output.vetted_code and not output.vetted_code.code.strip():
+                return "Pipeline approved but vetted_code.code is empty."
+        return super().check_output(output)
+
+
+# ---------------------------------------------------------------------------
+# Helpers
+# ---------------------------------------------------------------------------
+
+_SEV = {"LOW": 1, "MEDIUM": 2, "HIGH": 3}
```
sanzog03: "summary": "_SEV/_sev_rank duplicates the severity_rank mapping defined inside validator.py
  check_bandit", 

"failure_scenario": "Same {'LOW':1,'MEDIUM':2,'HIGH':3} mapping maintained in two files; a future severity change (or bandit adding a level) updated in
  one place silently desynchronizes pass/fail vs reported reasons"

akd_ext/agents/code_generator/schemas.py
```diff
@@ -0,0 +1,288 @@
...
+# ---------------------------------------------------------------------------
+
+
+class ValidationResult(BaseModel):
+    """Aggregated result from all three code validators."""
+
+    passed: bool
+    allowlist_passed: bool = True
+    bandit_passed: bool = True
+    intent_passed: bool = True
```
sanzog03: "summary": "Dead/redundant output state: ValidationResult.intent_passed is never set (always True), and total_attempts always equals len(attempt_history)",

"failure_scenario": "Callers reading intent_passed see True even when the intent checker rejected the code (its result lives elsewhere); total_attempts is a second copy of derivable state that can drift if the loop changes"

akd_ext/agents/closed_loop/cm1/agents.py
```diff
@@ -200,20 +233,711 @@ class CM1InterpretationPaperAssemblyAgent(InterpretationPaperAssemblyAgent):
...
+            generator_config=CodeGeneratorConfig(data_format_context=CM1_HARNESS_INTERFACE),
+            intent_checker_config=IntentCheckerConfig(data_format_context=CM1_HARNESS_INTERFACE),
+        ))
+        gen = await pipeline.arun(CodeGeneratorPipelineInputSchema(
+            context=self._design_context(params),
+            design_document=meta["design_document"],
+            success_criteria=meta.get("success_criteria", []),
+        ))
+        if isinstance(gen, CodeGeneratorPipelineOutputSchema) and gen.vetted_code:
+            self._module_path(params.job_id).write_text(gen.vetted_code.code, encoding="utf-8")
```
sanzog03: "summary": "The vetted analysis module is written to a local CWD file (cm1_analysis_module_{job_id}.py) that no code ever reads or submits, yet the user is told
'Plan approved and submitted.'",

"failure_scenario": "User approves the figure plan → _generate_module runs the pipeline and write_text()s the module locally, then returns 'Plan approved and
  submitted. Once the experiment completes … analyze.' No job_submit/MCP upload sends the module anywhere (grep finds only writes/exists-checks of _module_path, never a
  read). When the user later says 'analyze', the inherited base path fetches figures via job_plot from the remote server — produced independently of this module. Net
  effect: the generated module is dead output and the 'submitted' claim is false; if the harness was meant to run this module to plot, no figures will ever reflect it."
    },

akd_ext/agents/closed_loop/cm1/agents.py
```diff
@@ -200,20 +233,711 @@ class CM1InterpretationPaperAssemblyAgent(InterpretationPaperAssemblyAgent):
...
+        pending = self._pending_path(params.job_id)
+        base_ctx = self._design_context(params)
+        # The FIRST turn (no plan on disk yet) is the greeting, regardless of
+        # what the scientist typed to open the chat ("hi", "start", "show me").
+        is_fresh = not pending.exists()
+
+        if not is_fresh:
+            # Cumulative revision — give the designer the EXACT current plan
+            # and ask it to apply only this delta, keeping all else unchanged.
+            prev = json.loads(pending.read_text(encoding="utf-8"))
```
sanzog03: "summary": "json.loads(pending.read_text()) on the per-job state file is unguarded (also lines 854, 898); a partially-written/corrupt file raises JSONDecodeError that escapes _arun/_astream.",

"failure_scenario": "pending.write_text(json.dumps(...)) (line 682) is non-atomic. If a turn is interrupted mid-write, or two turns for the same job_id interleave,
  the JSON is truncated. The next 'design'/'approve' turn calls json.loads on it and raises JSONDecodeError uncaught — the whole chat turn crashes instead of falling back to a fresh plan. Same pattern in the approve branches of both _astream and _arun."

akd_ext/agents/closed_loop/cm1/agents.py
```diff
@@ -200,20 +233,711 @@ class CM1InterpretationPaperAssemblyAgent(InterpretationPaperAssemblyAgent):
...
+       ``ImageAnalyzer``) runs and returns the structured analysis.
+
+    ``run_context`` threads through the analyze branch unchanged so the
+    downstream paper-assembly stage receives the same context.
+    """
 
     config_schema = CM1ExperimentAnalysisConfig
+
+    @property
+    def _auth_headers(self) -> dict[str, str]:
```
sanzog03, line 458: "summary": "_auth_headers reimplements cm1/tools.py _mcp_auth_headers verbatim (CM1_MCP_AUTH → fastmcp.app → bearer vs x-api-key) instead of importing it.",

"failure_scenario": "Two copies of the auth-scheme resolution. A future change to one (e.g. a new scheme, or changing the default) silently desyncs them: job_submit (tools.py) would authenticate one way while job_status/job_plot (agents.py) use another, producing intermittent 401s that look like server problems. Import and call the existing helper."

akd_ext/agents/closed_loop/cm1/agents.py
rohit-sahoo: Fixed. The _generate_module now calls the code_submit MCP tool to ship the module (harness runs it, job_plot returns its figures); message only says "submitted" when that succeeds, else "generated, not submitted."

akd_ext/agents/closed_loop/cm1/agents.py
rohit-sahoo: Fixed, reads go through _read_pending (returns None on corrupt/missing instead of raising) and writes through _write_pending (atomic temp-file + os.replace); all three call sites updated.

akd_ext/agents/closed_loop/cm1/agents.py
rohit-sahoo: Fixed, imported and delegated: _auth_headers now just returns _mcp_auth_headers(self._api_key, self.config.mcp_url), single source of scheme resolution.

akd_ext/agents/code_generator/designer.py
rohit-sahoo: Fixed. pulled the assembly into _prompt.py (section/assemble_prompt) with a single DATA_FORMAT_CONTEXT_DESC; designer/generator/validator now call it instead of triplicating, and the divergent field docs are collapsed into the one constant.

akd_ext/agents/code_generator/validator.py
rohit-sahoo: Fixed. ImportFrom now flags relative imports (node.level > 0) as violations instead of skipping when node.module is None; the absolute-import allowlist check is unchanged. The plot module is standalone so any relative import is rejected up front rather than failing later with ImportError.

sanzog03 (approved):

